PR #17, opened 2015-03-05
Making fail fast mode for finagle configurable

dkwestbr:
I already had this code change pending so I figured I'd make a pull request for it.  I tried 0.3.4 but setting it to false actually made things worse (retries cause new peons to fill up middle managers), so at this time I think it is probably best to leave the default fail fast as True.

### DIFF
--- a/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
+++ b/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
@@ -295,7 +295,7 @@ object DruidBeams
 
             override def finagleHttpConnectionsPerHost = 2
 
-            override def finagleEnableFailedFastExceptions = true
+            override def finagleEnableFailFast = true
           },
           disco
         )
--- a/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
+++ b/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
@@ -294,6 +294,8 @@ object DruidBeams
             override def finagleHttpTimeout = 90.seconds
 
             override def finagleHttpConnectionsPerHost = 2
+
+            override def finagleEnableFailedFastExceptions = true
           },
           disco
         )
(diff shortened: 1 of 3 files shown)


gianm:
@dkwestbr: I'm +1 on this change, but can you give some more details around what was going on with your peons when fail-fast was off? Was it tranquility retrying task POSTs to your overlord? If so, I think we should fix that separately by making task submission idempotent. Finagle's fail-fast stuff will make doubling up on POSTs less likely but won't guarantee safety.

dkwestbr:
@gianm - I'll reproduce the issue to get some more details for you on what is going on.

gianm:
@dkwestbr Did you have a chance to try to reproduce the problem you ran into with fail-fast turned off?

If it does turn out to be retrying task POSTs, I think we can fix that by getting rid of the randomization of task ids. That'll make task creation idempotent. It'll also make replacing nuked tasks somewhat harder (usually doesn't come up in production, but can come up in testing). But we could put that randomization behind an option…

dkwestbr:
@gian I attempted to reproduce but unfortunately we couldn't reproduce.  I
believe this is because we had fixed the original problem causing the
FailedFastExceptions in the first place.  It appears to be due to using
iptables to filter traffic to the middle managers.  We tried re-enabling
the iptables to cause the issue again, but they had been recycled at that
point and we now seem to have difficulty reproducing even the original
FailedFastExceptions.

If they pop up again, I'll be sure to take advantage of the opportunity and
disable FailedFastExceptions and document observed behavior.

Although I wasn't able to reproduce, I did ask the engineer maintaining the
Druid cluster to describe what he had observed and he said it was as though
we had set replicants very high (we've yet to use more than 1 replicant).

At any rate, I think that making the task creation idempotent would be a
good thing.  I'll take a stab at it, unless you already had a way in mind
about how to best go about it.

gianm:
@dkwestbr The "as though we had set replicants very high" thing probably means that it was indeed POSTing the same task over and over. Ideally that should be idempotent. I think the best we can do with the current state of things in the overlord is to use the same task id for tasks that should be the same, which won't quite do what we want but will at least fail subsequent attempts to POST the same task. It'll fail them with a 400, which will cause tranquility to stop trying to POST the task, and then it should pick up the existing one on the next batch.

So, I'm going to merge this PR and then do that task id thing in a separate commit.

PR #20, opened 2015-03-31
Fixed druid community link

bobrik:


gianm:
:+1:

Can you please fill out the project CLA so we can accept this patch? It's here: http://druid.io/community/cla.html. Even though it says Druid on the page, the CLA actually covers all Metamarkets open source stuff. Thanks again!

bobrik:
I did that for druid-io/druid#1254 already, isn't that enough?

gianm:
Ah, yep, that is enough. Thanks!

PR #24, opened 2015-07-11
More granularities.

gianm:


src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -211,9 +211,16 @@ object DruidBeamMaker
...
+      case Granularity.MINUTE => tsUtc.minuteOfHour().get
+      case Granularity.FIVE_MINUTE => tsUtc.minuteOfHour().get
+      case Granularity.TEN_MINUTE => tsUtc.minuteOfHour().get
+      case Granularity.FIFTEEN_MINUTE => tsUtc.minuteOfHour().get
+      case Granularity.HOUR => tsUtc.hourOfDay().get
+      case Granularity.SIX_HOUR => tsUtc.hourOfDay().get
+      case Granularity.DAY => tsUtc.dayOfMonth().get
+      case Granularity.WEEK => tsUtc.weekOfWeekyear().get
+      case Granularity.MONTH => tsUtc.monthOfYear().get
+      case Granularity.YEAR => if (tsUtc.yearOfCentury().get == 100) 0 else tsUtc.yearOfCentury().get
```
xvrl: Is the special case needed. As far as I know, if we are using the default ISOChronology, yearOfCentury should always be between 0 and 99

gianm: Ok, good point. I was going based off the docs that said it could range up to 100, but there is a footnote that for ISOChronology that won't happen.

PR #29, opened 2015-09-03
Add contentType as parameter

nishantmonu51:
Add option to extend contentType. 
Can be used to send data to druid in smile format.

### DIFF
--- a/src/main/scala/com/metamx/tranquility/typeclass/JavaObjectWriter.scala
+++ b/src/main/scala/com/metamx/tranquility/typeclass/JavaObjectWriter.scala
@@ -32,4 +32,9 @@ trait JavaObjectWriter[A]
    * of objects.
    */
   def batchAsBytes(objects: java.util.Iterator[A]): Array[Byte]
+
+  /**
+   * @return content type of the serialized form
+   */
+  def contentType: String
 }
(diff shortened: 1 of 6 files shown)


src/main/scala/com/metamx/tranquility/typeclass/JavaObjectWriter.scala
```diff
@@ -32,4 +32,9 @@ trait JavaObjectWriter[A]
    * of objects.
    */
   def batchAsBytes(objects: java.util.Iterator[A]): Array[Byte]
+
+  /**
+   * @return content type of the serialized form
+   */
+  def contentType : String
```
gianm: should be,

```def contentType: String```

File: src/main/scala/com/metamx/tranquility/typeclass/ObjectWriter.scala [hunk 2]
```diff
@@ -33,6 +33,11 @@ trait ObjectWriter[A] extends Serializable
    * of objects.
    */
   def batchAsBytes(objects: TraversableOnce[A]): Array[Byte]
+
+  /**
+   * @return content type of the serialized form
+   */
+  def contentType : String
```
gianm: same spacing issue

gianm:
:+1: after formatting chages

PR #31, opened 2015-10-06
Update dependencies

gianm:
- Scala 2.10.5
- Druid 0.8.1
- Finagle 6.25.0
- Jackson 2.6.0
- scala-util 1.11.3
- slf4j 1.7.12

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ organization := "io.druid"
 
 name := "tranquility"
 
-scalaVersion := "2.10.4"
+scalaVersion := "2.10.5"
 
 lazy val root = project.in(file("."))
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -51,14 +51,14 @@ releaseSettings
 ReleaseKeys.publishArtifactsAction := PgpKeys.publishSigned.value
 
 val jacksonOneVersion = "1.9.13"
-val jacksonTwoVersion = "2.4.4"
-val druidVersion = "0.7.0"
-val finagleVersion = "6.24.0"
-val twitterUtilVersion = "6.23.0"
+val jacksonTwoVersion = "2.6.0"
```
xvrl: should we use latest 2.6.3 if we're already upgrading it?

gianm: sure

gianm: 2.6.1 is the latest jackson-module-scala, will use that one instead

gianm:
@xvrl updated jackson

xvrl:
:+1:

PR #32, opened 2015-10-09
Multi-project build: core, storm, samza.

gianm:
- Publish tranquility-core, tranquility-samza, tranquility-storm as separate artifacts
- Bump version to 0.6.0


fjy:
:+1:

build.sbt
```diff
@@ -57,43 +59,28 @@ val finagleVersion = "6.25.0"
...
-  "com.metamx" % "java-util" % "0.26.14" exclude("log4j", "log4j") force()
-)
-
-libraryDependencies ++= Seq(
-  "io.netty" % "netty" % "3.10.1.Final" force()
-)
-
-libraryDependencies ++= Seq(
+  "com.metamx" % "java-util" % "0.27.4" exclude("log4j", "log4j") force(),
+  "io.netty" % "netty" % "3.10.1.Final" force(),
```
xvrl: might as well update to 3.10.4 ?

build.sbt
```diff
@@ -30,18 +28,22 @@ pomExtra := (
...
       <developer>
         <name>Gian Merlino</name>
-        <organization>Metamarkets Group Inc.</organization>
-        <organizationUrl>https://www.metamarkets.com</organizationUrl>
+        <organization>Druid Project</organization>
+        <organizationUrl>http://druid.io/</organizationUrl>
       </developer>
     </developers>)
 
+parallelExecution in ThisBuild := false
```
xvrl: can we add some comments to explain why we need those options?

core/src/main/scala/com/metamx/tranquility/beam/ClusteredBeam.scala
```diff
@@ -1,18 +1,20 @@
 /*
- * Tranquility.
```
xvrl: can we put the header changes into a separate commit?

xvrl: mainly because it makes it possible to track file renames if we do it that way.

gianm: ok

build.sbt
gianm: sure

build.sbt
gianm: sure

gianm:
@xvrl pushed those changes

gianm:
new PR #33

PR #35, opened 2015-10-15
Use UTC DateTimes and Intervals whenever making decisions involving segments.

gianm:
This should make the time partitioning stuff work even if the user's timezone
is something other than UTC, or if the Timekeeper or Timestamper return
non-UTC times.


nishantmonu51:
+1

PR #36, opened 2015-10-15
Cross building with Travis

gianm:


xvrl:
@gianm can you use the same as here? https://github.com/metamx/scala-util/blob/master/.travis.yml
That should also make travis faster

gianm:
you mean adding sudo: false?

gianm:
ah I see, according to http://docs.travis-ci.com/user/workers/container-based-infrastructure/ that sends it to containers

gianm:
@xvrl added sudo: false

nishantmonu51:
+1

PR #49, opened 2015-11-07
Rollup-aware partitioning by default for Maps, and a Partitioner interface for other types.

### LINKED ISSUE #6: Default DruidBeams beamMergeFn should be rollup-aware
The default DruidBeams beam merger is just a HashPartitionBeam, which does it based on hashCode of the event type. This can lead to sub-optimal rollup because events with the same dimension values may end up in different druid partitions, just because they have different metrics.

Instead, the default should probably partition things based on dimensions and truncated timestamp, using the user-provided DruidRollup.

The workaround for this issue is to define a custom beamMergeFn, or to override the event type's hashCode to be the same for events that should roll up together.

gianm:
The default partitioning only works properly for Maps because other types are actually opaque
to tranquility (it doesn't know how the ObjectWriter is going to serialize them). But, you can
provide your own Partitioner (or an entire beamMergeFn, which you could already do).

Fixes #6.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/beam/MergingPartitioningBeam.scala
+++ b/core/src/main/scala/com/metamx/tranquility/beam/MergingPartitioningBeam.scala
@@ -19,11 +19,30 @@
 
 package com.metamx.tranquility.beam
 
-import com.metamx.tranquility.partition.HashCodePartitioner
+import com.metamx.common.scala.Logging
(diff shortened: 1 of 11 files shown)


fjy:
:+1:

fjy:
We should have a discussion about the voting requirements for Tranquility as it does not technically fall under Druid.

PR #54, opened 2015-11-23
Provide a default firehoseServicePattern. (fix #46)

### LINKED ISSUE #46: firehoseServicePattern should be optional
We could have a default. It should be based on the overlord service so things still work if people have the same named dataSource in two different druid clusters.

gianm:


fjy:
:+1:

Issue #56, opened 2015-11-30
Modify Beam trait so callers can tell which events were dropped


gianm:
Right now you know that events were dropped if the return value is less than the number of events you sent. But, it'd be good to know which events were dropped.

PR #57, opened 2015-12-02
Server and Tranquilizers.

### LINKED ISSUE #55: tranquility-server
This would be an HTTP server that runs a tranquility client. The idea is to provide a way for folks to use tranquility without embedding a Scala library in their app.

gianm:
This PR has a bunch of related stuff in it as separate commits.

Tranquilizer is a probably-easier-to-use API than Beam. It does batching and pipelining on its own and exposes two APIs: a future-oriented API where you can get the status of each individual message, and a "simple" API where you get overall stats and overall exceptions but not associated with any individual message.

BeamPacketizer is deprecated now as Tranquilizer is a replacement for it.

Tranquility Server is a server that you can send events to and have them forwarded along to Druid. It's a good way to use tranquility if you don't want to embed it or can't embed it (because you aren't using a JVM language).

Fixes #55


fjy:
:+1:

fjy:
Going with single +1s on tranquility until we reach some consensus on how to do reviews for Tranquility

xvrl:
@fjy we should use the same review policy as any other druid repository

gianm:
@xvrl @nishantmonu51 do either of you have time to review this? some of the things here (like the tranquilizer and the yaml config reading stuff) may also be useful for a tranquility-kafka module we want to add. would also like to look at #56 once this stuff is in master.

xvrl:
@gianm sure, I'll have a look, and hopefully @nishantmonu51 can do too. Do you mind adding an issue about tranquility-kafka so folks know it's being worked on? Thanks!

build.sbt
```diff
@@ -28,6 +31,8 @@ val coreDependencies = Seq(
   "com.twitter" %% "finagle-http" % finagleVersion,
   "org.slf4j" % "slf4j-api" % "1.7.12" force() force(),
   "org.slf4j" % "jul-to-slf4j" % "1.7.12" force() force(),
+  "org.apache.httpcomponents" % "httpclient" % apacheHttpVersion,
+  "org.apache.httpcomponents" % "httpcore" % apacheHttpVersion,
```
xvrl: any reason why we are adding those here?

build.sbt
```diff
@@ -65,6 +70,15 @@ val coreDependencies = Seq(
   "javax.validation" % "validation-api" % "1.1.0.Final" force()
 )
 
+val loggingDependencies = Seq(
+  "ch.qos.logback" % "logback-core" % "1.1.2",
+  "ch.qos.logback" % "logback-classic" % "1.1.2",
```
xvrl: is logback only necessary for the server? If so we might not want to put it into core tranquility

gianm: I think it makes sense to only have it in the server. In core tranquility it often conflicts with other loggers people want to use (as tranquility is always embedded in some larger app).

But, server is its own standalone app, so it needs a logger.

build.sbt
gianm: One of the new dependencies pulled in different versions that caused conflicts. I didn't track down which one, just adjusted this to pull in a particular version.

gianm:
@xvrl good point. #59

fjy:
@xvrl not everyone understands scala or cares about reviewing tranquility the same way they do with druid

gianm:
added some docs to the PR - https://github.com/gianm/tranquility/blob/server/docs/server.md

build.sbt
xvrl: maybe add a comment to explain?

build.sbt
xvrl: ah I didn't realize it was only embedding logging for test in core, I misread

docs/server.md
```diff
@@ -0,0 +1,98 @@
...
+### Other properties
+
+|Property|Description|Default|
+|--------|-----------|-------|
+|`druid.indexService`|Druid Overlord's service.name|druid/overlord|
+|`druid.partitions`|Number of Druid partitions to create.|1|
+|`druid.replicants`|Number of instances of each Druid partition to create. This is the *total* number of instances, so 2 replicants means 2 tasks will be created.|1|
+|`zookeeper.connect`|ZooKeeper connect string.|none; must be provided|
+|`zookeeper.timeout`|ZooKeeper session timeout. ISO8601 duration.|PT20S|
+|`disco.path`|Curator service discovery path.|/druid/discovery|
```
xvrl: should we use the same property name as druid here?

server/src/main/scala/com/metamx/tranquility/server/http/TranquilityServlet.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.jboss.netty.handler.codec.http.HttpResponseStatus
+import org.scalatra.ScalatraServlet
+import scala.collection.mutable
+
+class TranquilityServlet(
+  tranquilizers: Map[String, Tranquilizer[Dict]]
+) extends ScalatraServlet with Logging
+{
+  get("/") {
+    contentType = "text/plain"
```
xvrl: hey, that was my idea :)

gianm: good idea!

docs/server.md
gianm: sure, that sounds good. I'll also change `druid.indexService` to the matching druid property (`druid.selectors.indexing.serviceName` I think)

xvrl: agree, ideally anything that already exists in Druid should have the same property name. I think ZK  is also different here?

gianm: I was less sure about that one as it's called `druid.zk.service.host`, but the zk cluster is not just used for druid stuff- it's also used for internal tranquillity coordination. So I was concerned it would be misleading?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+            } else {
+              lock.wait(math.max(1, bufferStartMillis + lingerMillis - System.currentTimeMillis()))
+            }
+          }
+          if (log.isDebugEnabled) {
+            log.debug(s"Buffer with ${buffer.size} messages has lingered too long, sending.")
+          }
+
+          assert(buffer.nonEmpty)
+          assert(pendingBatches < maxPendingBatches)
```
xvrl: aren't those two conditions guaranteed by the while loop above?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+  lingerMillis: Long
+) extends Logging
+{
+  require(maxBatchSize >= 0, "batchSize >= 0")
+  require(maxPendingBatches >= 1, "maxPendingBatches >= 1")
+  require(lingerMillis >= 0, "lingerMillis >= 0")
+
+  @volatile private var started: Boolean = false
+
+  case class MessageHolder(Message: MessageType, future: Promise[Unit])
```
xvrl: should this be lowercase `message`?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: yes, they are, that's why they're asserts :)

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: whoops, yes. fixed

docs/server.md
xvrl: ok that's faire we can keep it as is.

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
```diff
@@ -76,7 +76,86 @@ import scala.language.reflectiveCalls
   */
 object DruidBeams
 {
-  private val DefaultTimestampSpec = new TimestampSpec("timestamp", "iso", null)
+  val DefaultTimestampSpec = new TimestampSpec("timestamp", "iso", null)
```
nishantmonu51: why not private ? who uses it ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
```diff
@@ -0,0 +1,148 @@
...
+  private val failed   = new AtomicLong
+
+  /**
+    * Send a message. The actual send may happen asynchronously. If you must know whether or not there was an
+    * exception sending it, call "flush", which will flush all pending messages and throw an exception if there
+    * was one.
+    *
+    * This method may throw an exception if there was one, although it is not guaranteed. Furthermore, if this
+    * method does throw an exception, it might be for a different message.
+    */
```
nishantmonu51: this seems confusing, I wonder how would someone handle the exceptions from this message ? 
e.g if there is an exception  
   -  do we retry sending the same message again as the exception can be from prev message ? 
   -  what should we do in case retry also fails with exception from some other prev message ? 
   -  is it ok to miss exceptions when there are multiple message send failures between two send calls ? 
  
I wonder if it would be less confusing to accept messages async always without any exception and always report exceptions in flush method ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
```diff
@@ -0,0 +1,148 @@
...
+  def send(message: MessageType) {
+    maybeThrow()
+    received.incrementAndGet()
+    pending.incrementAndGet()
+    tranquilizer.send(message) respond { res =>
```
nishantmonu51: is send guaranteed to never throw exception and always return a future that will have exception in it ? 
In case there will be any exception in send the counters will be messed up and the adapter will be unusable.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+    * MessageDroppedException, which means that a message was dropped due to unrecoverable reasons. With Druid this
+    * can be caused by message timestamps being outside the configured windowPeriod.
+    *
+    * NB: This can actually be wrong: if *any* message in a batch of messages is dropped, then all messages in that
+    * batch will be marked as dropped (even if they actually made it out). This is a limitation of how Tranquility
+    * tracks which messages were and were not dropped. This could be improved once
+    * https://github.com/druid-io/tranquility/issues/56 is addressed.
+    */
+  def send(message: MessageType): Future[Unit] = {
+    requireStarted()
```
nishantmonu51: this throws ISE, should we instead return a Future having the ISE ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+
+  // Swap out the current buffer, returning it and incrementing pendingBatches. Will block if more than
+  // maxPendingBatches are already pending.
+  private def swap(): Buffer[MessageHolder] = {
+    lock.synchronized {
+      while (pendingBatches >= maxPendingBatches) {
+        if (log.isDebugEnabled) {
+          log.debug(s"pendingBatches[$pendingBatches] >= maxPendingBatches[$maxPendingBatches], waiting...")
+        }
+        lock.wait()
```
nishantmonu51: Its possible for multiple threads to wait here, after this all of them can start and swap the buffer, 
wondering if this can possibly lead to a case when a batch has 0 messages to send ? 
do we need any handling for sending empty buffer?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+  * Exception indicating that a message was dropped "on purpose" by the beam. This is not a recoverable exception
+  * and so the message must be discarded.
+  */
+class MessageDroppedException(message: String) extends Exception(message)
+
+object Tranquilizer
+{
+  val DefaultMaxBatchSize      = 2000
```
nishantmonu51: do we need to align the defaults with default firehoseChunkSize ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+  val DefaultMaxPendingBatches = 5
+  val DefaultLingerMillis      = 0
+  val DefaultSendThreads       = math.max(1, Runtime.getRuntime.availableProcessors() / 2)
```
nishantmonu51: where is this used ? don't we have a single send thread always ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,310 @@
...
+/**
+  * Exception indicating that a message was dropped "on purpose" by the beam. This is not a recoverable exception
+  * and so the message must be discarded.
+  */
+class MessageDroppedException(message: String) extends Exception(message)
+
+object Tranquilizer
+{
+  val DefaultMaxBatchSize      = 2000
+  val DefaultMaxPendingBatches = 5
```
nishantmonu51: just curious how did you came to these defaults for batchSize and pendingBatches? 
are they based on any properties on underlying beam ?

docs/server.md
```diff
@@ -0,0 +1,98 @@
...
+
+```bash
+bin/tranquility-server conf/server.yaml
+```
+
+## Configuration reference
+
+With the exception of the global properties (which must be provided at the top level), most properties can be provided
+at the top level or per-dataSource. If the same property is provided at both levels, the per-dataSource version will
+take predence. This provides a way to set default properties that apply to all dataSources, while still allowing
```
nishantmonu51: s/predence/precedence/g

server/src/main/scala/com/metamx/tranquility/server/http/TranquilityServlet.scala
nishantmonu51: Nice.

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
gianm: The idea with the simple adapter is that if you got any exceptions you should probably abort the entire process and discard the simple adapter, because the beams are doing a lot of retrying internally and anything that makes it out is fatal.

- In storm that means the bolt dies and is respawned by the storm supervisor
- In samza that means the container dies and is respawned by the samza app master
- In server that means the request fails

The idea with throwing them from "send" is that if something bad happens, you can fail fast without waiting for the next flush, which might be a long way away.

I can update the doc comment to reflect that expected usage if it sounds good.

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
gianm: yes, this is guaranteed except in case of:

- the tranquilizer not being started yet, where you get an immediate ISE
- OOME, but if that happens I think it makes sense to abort the program and allow higher-level supervisor to restart it, since OOMEs often cause bad states to exist

gianm: also, the expected use case is that if "send" or "flush" throws an exception then you should stop using the adapter.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: hmm, I think ISE is ok, as you're breaking the contract of this class if you call send before start.

gianm: do you see this causing problems?

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
gianm: oops, nobody, that was from a previous version of the patch… will make private again

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: nowhere, removing

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: nope, these were the defaults from BeamPacketizer; I just made them up…

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: yes, I think this is a problem, will look into it.

docs/server.md
gianm: will fix

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: @nishantmonu51 pushed with changes

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,302 @@
...
+
+      if (buffer.isEmpty) {
+        bufferStartMillis = System.currentTimeMillis()
+        lock.notifyAll()
+      }
+
+      buffer += holder
+
+      val _exBuffer: Option[Buffer[MessageHolder]] = if (buffer.size == maxBatchSize) {
+        Some(swap())
```
xvrl: are we guaranteed `pendingBatches < maxPendingBatches` here?

gianm: yes, because if `buffer.size == maxBatchSize` at this point, then the earlier while loop would have waited for `pendingBatches < maxPendingBatches`

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,302 @@
...
+    if (!started) {
+      throw new IllegalStateException("Not started")
+    }
+  }
+
+  // Swap out the current buffer immediately, returning it and incrementing pendingBatches.
+  // Must be called while holding the "lock".
+  // Preconditions: buffer must be non-empty and pendingBatches must be lower than maxPendingBatches.
+  private def swap(): Buffer[MessageHolder] = {
+    assert(buffer.nonEmpty && buffer.size <= maxBatchSize)
```
xvrl: is there any harm in wrapping this method in lock.synchronized for good measure? then the only requirement when calling swap would be to have the preconditions set.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,302 @@
...
+  private val sendThread = loggingThread {
+    try {
+      while (!Thread.currentThread().isInterrupted) {
+        val exBuffer = lock.synchronized {
+          while (
+            buffer.isEmpty ||
+              pendingBatches >= maxPendingBatches ||
+              (lingerMillis > 0 && System.currentTimeMillis() < bufferStartMillis + lingerMillis)
+          ) {
+            if (lingerMillis == 0) {
```
xvrl: minor optimization, you could add `|| bufferStartMillis == 0` here to avoid waking the thread every milliseconds on empty buffers.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: I think there's no gain in doing that, since having the lock is the only reasonable way of being sure that the preconditions are true, so callers must have the lock anyway. If they don't, it's probably good that this will throw an exception on the `lock.notifyAll` so it will be clear something is buggy.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,302 @@
...
+  require(lingerMillis >= 0, "lingerMillis >= 0")
+
+  @volatile private var started: Boolean = false
+
+  case class MessageHolder(message: MessageType, future: Promise[Unit])
+
+  private val lock             : AnyRef                = new AnyRef
+  private var buffer           : Buffer[MessageHolder] = new ArrayBuffer[MessageHolder]
+  private var bufferStartMillis: Long                  = 0L
+  private var pendingBatches   : Int                   = 0
```
xvrl: can we add a comment to explain upon updating which variables we are supposed to call lock.notifyAll ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: good idea, adding

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: sure

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
nishantmonu51: makes sense, lets make it clear in the doc.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
nishantmonu51: was wondering about the behavior of the simple adapter with this as it will mess up with the counters that it maintains., 
I am fine with the exception itself, lets also mention it in javadoc.

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
nishantmonu51: I wonder if there can be any recoverable errors thrown by the beams, like the druid tasks failed but the underlying beam created next set of tasks later and recovered ?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -0,0 +1,305 @@
...
+      if (buffer.isEmpty) {
+        bufferStartMillis = System.currentTimeMillis()
+        lock.notifyAll()
+      }
+
+      buffer += holder
+
+      val _exBuffer: Option[Buffer[MessageHolder]] = if (buffer.size == maxBatchSize) {
+        Some(swap())
+      } else if (lingerMillis == 0 && pendingBatches < maxPendingBatches) {
```
nishantmonu51: can remove the pendingBatches check here also.

core/src/main/scala/com/metamx/tranquility/tranquilizer/SimpleTranquilizerAdapter.scala
gianm: Currently the way it works (and I think it should keep working like this) is that the beams internally do a bunch of retries, so they don't throw recoverable errors. If they decide to give up they just report the message as dropped.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: I don't think we can, because at this point buffer.size < maxBatchSize, which means pendingBatches could potentially be >= maxPendingBatches. And we have to check for that, as you are not allowed to swap unless there is room (pendingBatches < maxPendingBatches).

xvrl:
:+1: looks good to me once scala 2.10 build passes.

PR #65, opened 2015-12-09
tranquility-kafka module

dclim:
This PR adds a module to Tranquility to help push data from Kafka into the indexing service through Tranquility. Multiple instances of tranquility-kafka can be run at the same time to provide scalability and high availability, and an instance can read from multiple Kafka topics and write to multiple Druid datasources. Tranquility-kafka supports regex pattern based mapping of Kafka topic to a Druid ingestion spec, thus allowing indexing tasks for new datasources to be automatically created when a new Kafka topic is created that matches the regex. If additional Kafka topic -> ingestion spec mappings need to be added, tranquility-kafka supports rolling updates without data loss.

To build, run: `sbt universal:packageZipTarball`

Unpack the tgz in `/kafka/target/universal/tranquility-kafka-x.x.x.tgz` and run `bin/tranquility-kafka -f conf/config.properties` to start. You'll need to modify config.properties and add some ingestion specs to do anything useful.

I'll be adding unit tests and documentation, but wanted to put this out to get some feedback.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,8 @@ val sparkVersion = "1.6.0"
 val scalatraVersion = "2.3.1"
 val jettyVersion = "9.2.5.v20141112"
 val apacheHttpVersion = "4.3.3"
+val kafkaVersion = "0.9.0.0"
+val airlineVersion = "0.7"
 
 def dependOnDruid(artifact: String) = {
   ("io.druid" % artifact % druidVersion
@@ -100,13 +102,21 @@ val serverDependencies = Seq(
   "org.eclipse.jetty" % "jetty-servlet" % jettyVersion
 ) ++ loggingDependencies
 
+val kafkaDependencies = Seq(
+  "org.apache.kafka" %% "kafka" % kafkaVersion
+    exclude("org.slf4j", "slf4j-log4j12")
+    exclude("log4j", "log4j")
+    force(),
+  "io.airlift" % "airline" % airlineVersion
+) ++ loggingDependencies
+
 val coreTestDependencies = Seq(
   "org.scalatest" %% "scalatest" % "2.2.5" % "test",
   dependOnDruid("druid-services") % "test",
   "org.apache.curator" % "curator-test" % "2.6.0" % "test" exclude("log4j", "log4j") force(),
   "com.sun.jersey" % "jersey-servlet" % "1.17.1" % "test" force(),
-  "junit" % "junit" % "4.11" % "test",
-  "com.novocode" % "junit-interface" % "0.11-RC1" % "test",
+  "junit" % "junit" % "4.12" % "test",
+  "com.novocode" % "junit-interface" % "0.11" % "test",
   "ch.qos.logback" % "logback-core" % "1.1.2" % "test",
   "ch.qos.logback" % "logback-classic" % "1.1.2" % "test",
   "org.apache.logging.log4j" % "log4j-to-slf4j" % "2.4" % "test",
@@ -123,6 +133,10 @@ val samzaTestDependencies = Seq(
 
 val serverTestDependencies = Seq(
   "org.scalatra" %% "scalatra-test" % scalatraVersion % "test"
+)
+
+val kafkaTestDependencies = Seq(
+  "org.easymock" % "easymock" % "3.4" % "test"
 )
 
 lazy val commonSettings = Seq(
@@ -164,7 +178,7 @@ lazy val commonSettings = Seq(
 lazy val root = project.in(file("."))
   .settings(commonSettings: _*)
   .settings(publishArtifact := false)
-  .aggregate(core, storm, samza, spark, server)
+  .aggregate(core, storm, samza, spark, server, kafka)
 
 lazy val core = project.in(file("core"))
   .settings(commonSettings: _*)
@@ -201,6 +215,15 @@ lazy val server = project.in(file("server"))
   .settings(name := "tranquility-server")
   .settings(libraryDependencies ++= (serverDependencies ++ serverTestDependencies))
   .settings(publishArtifact in Test := false)
+  .dependsOn(core % "test->test;compile->compile")
+
+lazy val kafka = project.in(file("kafka"))
+  .settings(commonSettings: _*)
+  .settings(name := "tranquility-kafka")
+  .settings(libraryDependencies ++= (kafkaDependencies ++ kafkaTestDependencies))
+  .settings(publishArtifact in Test := false)
+  .settings(javaOptions in Universal ++= Seq("-J-Dfile.encoding=UTF-8", "-J-Duser.timezone=UTC"))
+  .enablePlugins(JavaAppPackaging)
   .dependsOn(core % "test->test;compile->compile")
 
 lazy val distribution = project.in(file("distribution"))
--- a/core/src/main/scala/com/metamx/tranquility/config/ConfigHelper.scala
+++ b/core/src/main/scala/com/metamx/tranquility/config/ConfigHelper.scala
@@ -1,0 +1,143 @@
+/*
+ * Licensed to Metamarkets Group Inc. (Metamarkets) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Metamarkets licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.metamx.tranquility.config
+
+import com.metamx.common.scala.Predef._
+import com.metamx.common.scala.Yaml
+import com.metamx.common.scala.untyped._
+import com.metamx.tranquility.druid.DruidBeams
+import com.metamx.tranquility.druid.DruidEnvironment
+import com.metamx.tranquility.druid.DruidGuicer
+import com.metamx.tranquility.druid.DruidLocation
+import com.metamx.tranquility.finagle.FinagleRegistry
+import com.metamx.tranquility.tranquilizer.Tranquilizer
+import io.druid.segment.realtime.FireDepartment
+import java.io.InputStream
+import java.util.Properties
+import org.apache.curator.framework.CuratorFramework
+import org.skife.config.ConfigurationObjectFactory
+
+/**
+ * Helper methods to perform shared tasks based on common configuration file semantics.
+ */
+object ConfigHelper
+{
+  /**
+   * Builds configuration object from YAML file.
+   *
+   * This method may have the side effect of modifying system properties if the YAML file contains properties beginning
+   * with "druid.extensions." - this is necessary as DruidGuicer reads the system properties to know what extensions
+   * need to be loaded. If the properties have already been set elsewhere, the system properties will not be modified.
+   */
+  def readConfigYaml[T <: TranquilityConfig](
+    in: InputStream,
+    clazz: Class[T]
+    ): (T, Map[String, DataSourceConfig[T]], Dict) =
+  {
+    val yaml = in.withFinally(_.close()) { in =>
+      dict(Yaml.load(in))
+    }
+    val globalProperties: Dict = getAsDict(yaml, "properties")
+    val globalConfig = mapConfig(globalProperties, clazz)
+
+    for (prop <- globalProperties.filterKeys(_.startsWith("druid.extensions.")))
+    {
+      if (!sys.props.contains(prop._1)) {
+        // copy properties to system as DruidGuicer reads system properties to know what extensions need to be loaded
+        sys.props += (prop._1 -> prop._2.toString)
+      }
+    }
+
+    val dataSources: Dict = getAsDict(yaml, "dataSources")
+    val dataSourceConfigs = for ((dataSource, x) <- dataSources) yield {
+      val d = dict(x)
+      val dataSourceProperties = getAsDict(d, "properties")
+      for (k <- globalConfig.GlobalProperties) {
+        require(!dataSourceProperties.contains(k), s"Property '$k' cannot be per-dataSource (found in[$dataSource]).")
+      }
+
+      val dataSourceSpec = getAsDict(d, "spec")
+      val fireDepartment = DruidGuicer.objectMapper.convertValue(normalizeJava(dataSourceSpec), classOf[FireDepartment])
+      val config = mapConfig(globalProperties ++ dataSourceProperties, clazz)
+
+      // Sanity check: two ways of providing dataSource, they must match
+      require(
+        dataSource == fireDepartment.getDataSchema.getDataSource,
+        s"dataSource[$dataSource] did not match spec[${fireDepartment.getDataSchema.getDataSource}]"
+      )
+
+      (dataSource, DataSourceConfig(config, fireDepartment))
+    }
+
+    (globalConfig, dataSourceConfigs, globalProperties)
+  }
+
+  private def mapConfig[A](d: Dict, clazz: Class[A]): A = {
+    val properties = new Properties
+    for ((k, v) <- d if v != null) {
+      properties.setProperty(k, String.valueOf(v))
+    }
+    val configFactory = new ConfigurationObjectFactory(properties)
+    configFactory.build(clazz)
+  }
+
+  private def getAsDict(d: Dict, k: String): Dict = {
+    Option(d.getOrElse(k, null)).map(dict(_)).getOrElse(Dict())
+  }
+
+  def createTranquilizer[T <: TranquilityConfig](
+    dataSourceConfig: DataSourceConfig[T],
+    finagleRegistry: FinagleRegistry,
+    curator: CuratorFramework
+    ): Tranquilizer[Dict] =
+  {
+    createTranquilizerWithLocation(dataSourceConfig, finagleRegistry, curator, null)
+  }
+
+  def createTranquilizerWithLocation[T <: TranquilityConfig](
+    dataSourceConfig: DataSourceConfig[T],
+    finagleRegistry: FinagleRegistry,
+    curator: CuratorFramework,
+    location: DruidLocation
+    ): Tranquilizer[Dict] =
+  {
+    val environment = DruidEnvironment(dataSourceConfig.config.druidIndexingServiceName)
+
+    // Create beam for this dataSource
+    var beamBuilder = DruidBeams
+      .builder(environment, dataSourceConfig.fireDepartment)
+      .curator(curator)
+      .finagleRegistry(finagleRegistry)
+      .partitions(dataSourceConfig.config.taskPartitions)
+      .replicants(dataSourceConfig.config.taskReplicants)
+      .druidBeamConfig(dataSourceConfig.config.druidBeamConfig)
+
+    if (location != null) {
+      beamBuilder = beamBuilder.location(location)
+    }
+
+    Tranquilizer(
+      beamBuilder.buildBeam(),
+      dataSourceConfig.config.tranquilityMaxBatchSize,
+      dataSourceConfig.config.tranquilityMaxPendingBatches,
+      dataSourceConfig.config.tranquilityLingerMillis
+    )
+  }
+}
--- a/core/src/main/scala/com/metamx/tranquility/config/config.scala
+++ b/core/src/main/scala/com/metamx/tranquility/config/config.scala
@@ -1,0 +1,103 @@
+/*
+ * Licensed to Metamarkets Group Inc. (Metamarkets) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  Metamarkets licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.metamx.tranquility.config
+
+import com.metamx.common.scala.net.curator.DiscoAnnounceConfig
+import com.metamx.common.scala.net.curator.DiscoConfig
+import com.metamx.tranquility.beam.ClusteredBeamTuning
+import com.metamx.tranquility.druid.DruidBeamConfig
+import com.metamx.tranquility.tranquilizer.Tranquilizer
+import io.druid.segment.realtime.FireDepartment
+import org.joda.time.Period
+import org.skife.config.Config
+
+abstract class TranquilityConfig(globalProperties: Set[String]) extends DiscoConfig
+{
+  def this() {
+    this(Set[String]())
+  }
+
+  val GlobalProperties = Set[String]() ++ globalProperties
+
+  @Config(Array("druid.selectors.indexing.serviceName"))
+  def druidIndexingServiceName: String = "druid/overlord"
+
+  @Config(Array("task.partitions"))
+  def taskPartitions: Int = ClusteredBeamTuning().partitions
+
+  @Config(Array("task.replicants"))
+  def taskReplicants: Int = ClusteredBeamTuning().replicants
+
+  @Config(Array("task.warmingPeriod"))
+  def taskWarmingPeriod: Period = ClusteredBeamTuning().warmingPeriod
+
+  @Config(Array("zookeeper.connect"))
+  def zookeeperConnect: String
+
+  @Config(Array("zookeeper.timeout"))
+  def zookeeperTimeout: Period = new Period("PT20S")
+
+  @Config(Array("tranquility.maxBatchSize"))
+  def tranquilityMaxBatchSize: Int = Tranquilizer.DefaultMaxBatchSize
+
+  @Config(Array("tranquility.maxPendingBatches"))
+  def tranquilityMaxPendingBatches: Int = Tranquilizer.DefaultMaxPendingBatches
+
+  @Config(Array("tranquility.lingerMillis"))
+  def tranquilityLingerMillis: Int = Tranquilizer.DefaultLingerMillis
+
+  @Config(Array("druid.discovery.curator.path"))
+  def discoPath: String = "/druid/discovery"
+
+  @Config(Array("druidBeam.firehoseGracePeriod"))
+  def firehoseGracePeriod: Period = DruidBeamConfig().firehoseGracePeriod
+
+  @Config(Array("druidBeam.firehoseQuietPeriod"))
+  def firehoseQuietPeriod: Period = DruidBeamConfig().firehoseQuietPeriod
+
+  @Config(Array("druidBeam.firehoseRetryPeriod"))
+  def firehoseRetryPeriod: Period = DruidBeamConfig().firehoseRetryPeriod
+
+  @Config(Array("druidBeam.firehoseChunkSize"))
+  def firehoseChunkSize: Int = DruidBeamConfig().firehoseChunkSize
+
+  @Config(Array("druidBeam.randomizeTaskId"))
+  def randomizeTaskId: Boolean = DruidBeamConfig().randomizeTaskId
+
+  @Config(Array("druidBeam.indexRetryPeriod"))
+  def indexRetryPeriod: Period = DruidBeamConfig().indexRetryPeriod
+
+  @Config(Array("druidBeam.firehoseBufferSize"))
+  def firehoseBufferSize: Int = DruidBeamConfig().firehoseBufferSize
+
+  def druidBeamConfig: DruidBeamConfig = DruidBeamConfig.builder()
+    .firehoseGracePeriod(firehoseGracePeriod)
+    .firehoseQuietPeriod(firehoseQuietPeriod)
+    .firehoseRetryPeriod(firehoseRetryPeriod)
+    .firehoseChunkSize(firehoseChunkSize)
+    .randomizeTaskId(randomizeTaskId)
+    .indexRetryPeriod(indexRetryPeriod)
+    .firehoseBufferSize(firehoseBufferSize)
+    .build()
+
+  override def discoAnnounce: Option[DiscoAnnounceConfig] = None
+}
+
+case class DataSourceConfig[T <: TranquilityConfig](config: T, fireDepartment: FireDepartment)
--- a/docs/kafka.md
+++ b/docs/kafka.md
@@ -1,0 +1,147 @@
+# Kafka
+
+Tranquility Kafka is an application which simplifies the ingestion of data from Kafka. It is scalable and highly
+available through the use of Kafka partitions and consumer groups, and can be configured to push data from multiple
+Kafka topics into multiple Druid dataSources.
+
+## Setup
+
+### Getting Tranquility Kafka
+
+Tranquility Kafka currently must be built from source. To build, run `sbt universal:packageZipTarball`. The build output
+will be contained in the archive `/kafka/target/universal/tranquility-kafka-x.x.x.tgz`. Unpack this tarball to the
+directory of choice by running `tar -xzf tranquility-kafka-x-x-x.tgz`.
+
+### Configuration
+
+Tranquility Kafka uses a YAML file for configuration. You can see an example in `conf/tranquility-kafka.yaml.example` of the
+tarball distribution. You can start off your installation by copying this example file to `conf/tranquility-kafka.yaml`. The
+YAML file has two sections:
+
+1. `dataSources` - per dataSource configuration.
+2. `properties` - general properties that apply to all dataSources. See "Configuration reference" for details.
+
+The dataSources key should contain a mapping of dataSource name to configuration. Each dataSource configuration
+has two sections:
+
+1. `spec` - a YAML representation of a [Druid ingestion spec](http://druid.io/docs/latest/ingestion/index.html). Note
+that the `ioConfig` is expected to be of type `realtime` but otherwise empty (null firehose, null plumber). This is
+because Tranquility supplies its own firehose and plumber.
+2. `properties` - per dataSource properties. See "Configuration reference" for details.
+
+### Running
+
+If you've saved your configuration into `conf/tranquility-kafka.yaml`, run the application with:
+
+```bash
+bin/tranquility-kafka -configFile conf/tranquility-kafka.yaml
+```
+
+## Configuration reference
+
+With the exception of the global properties (which must be provided at the top level), most properties can be provided
+at the top level or per-dataSource. If the same property is provided at both levels, the per-dataSource version will
+take precedence. This provides a way to set default properties that apply to all dataSources, while still allowing
+customization for certain dataSources.
+
+### Global properties
+
+|Property|Description|Default|
+|--------|-----------|-------|
+|`kafka.zookeeper.connect`|ZooKeeper connect string for Kafka.|none; must be provided|
+|`kafka.group.id`|Group ID for Kafka consumers. **Note:** This must be the same for all instances to avoid data duplication!|tranquility-kafka|
+|`consumer.numThreads`|The number of threads that will be made available to the Kafka consumer for fetching messages.|{numProcessors} - 1|
+|`commit.periodMillis`|The frequency with which consumer offsets will be committed to ZooKeeper to track processed messages.|15000|
+|`kafka.*`|Any properties that begin with *kafka.* will be passed to the underlying Kafka consumer with the *kafka.* prefix removed. For example, if you set `kafka.consumer.id=myConsumer`, the Kafka consumer will be passed the property `consumer.id=myConsumer`. Note that Tranquility Kafka requires `auto.commit.enable` to be *false* and any attempt to set this property will be overridden. |none|
+|`druid.extensions.*`|You may load Druid extensions into Tranquility Kafka using the extension properties defined on the Druid [common configuration](http://druid.io/docs/latest/configuration/index.html) page. This is required if your dataSource configuration contains fields not part of core Druid such as *approxHistogram*. As an example, to load the histogram extension, set `druid.extensions.coordinates: "[\"io.druid.extensions:druid-histogram:x.x.x\"]"`.|none|
+
+### Other properties
+
+|Property|Description|Default|
+|--------|-----------|-------|
+|`druid.discovery.curator.path`|Curator service discovery path.|/druid/discovery|
+|`druid.selectors.indexing.serviceName`|The druid.service name of the indexing service Overlord node.|druid/overlord|
+|`topicPattern`|A regular expression used to match Kafka topics to dataSource configurations. See "Matching Topics to Data Sources" for details.|{match nothing}, must be provided|
+|`topicPattern.priority`|If multiple topicPatterns match the same topic name, the highest priority dataSource configuration will be used. A higher number indicates a higher priority. See "Matching Topics to Data Sources" for details.|1|
+|`useTopicAsDataSource`|Use the Kafka topic as the dataSource name instead of the one provided in the YAML. Useful when combined with a topicPattern that matches more than one Kafka topic. See "Matching Topics to Data Sources" for details.|false|
+|`reportDropsAsExceptions`|Whether or not dropped messages will cause an exception and terminate the application.|false|
+|`task.partitions`|Number of Druid partitions to create.|1|
+|`task.replicants`|Number of instances of each Druid partition to create. This is the *total* number of instances, so 2 replicants means 2 tasks will be created.|1|
+|`task.warmingPeriod`|If nonzero, create Druid tasks early. This can be useful if tasks take a long time to start up in your environment.|PT0M|
+|`zookeeper.connect`|ZooKeeper connect string.|none; must be provided|
+|`zookeeper.timeout`|ZooKeeper session timeout. ISO8601 duration.|PT20S|
+|`tranquility.maxBatchSize`|Maximum number of messages to send at once.|2000|
+|`tranquility.maxPendingBatches`|Maximum number of batches that may be in flight before we block and wait for one to finish.|5|
+|`tranquility.lingerMillis`|Wait this long for batches to collect more messages (up to maxBatchSize) before sending them. Set to zero to disable waiting.|0|
+|`druidBeam.firehoseGracePeriod`|Druid indexing tasks will shut down this long after the windowPeriod has elapsed.|PT5M|
+|`druidBeam.firehoseQuietPeriod`|Wait this long for a task to appear before complaining that it cannot be found.|PT1M|
+|`druidBeam.firehoseRetryPeriod`|Retry for this long before complaining that events could not be pushed|PT1M|
+|`druidBeam.firehoseChunkSize`|Maximum number of events to send to Druid in one HTTP request.|1000|
+|`druidBeam.randomizeTaskId`|True if we should add a random suffix to Druid task IDs. This is useful for testing.|false|
+|`druidBeam.indexRetryPeriod`|If an indexing service overlord call fails for some apparently-transient reason, retry for this long before giving up.|PT1M|
+|`druidBeam.firehoseBufferSize`|Size of buffer used by firehose to store events.|100000|
+
+
+## Matching Topics to Data Sources
+
+Tranquility Kafka supports matching multiple Kafka topics to Druid dataSources through the use of the regex-based
+`topicPattern` property. A topic can be mapped to at most one dataSource at a time; if there are multiple topic patterns
+that match a given Kafka topic, the one with the highest-valued `topicPattern.priority` will be used.
+
+This can be combined with the `useTopicAsDataSource` property to map multiple topics to multiple similar dataSources
+without having to add repeated configurations for each dataSource. When `useTopicAsDataSource` is set to `true`, the
+dataSource name for this entry will be ignored and the Kafka topic will be used instead. Since the `topicPattern`
+matches using regular expressions, this can be used to watch for Kafka topics that did not exist when the application
+was started and dynamically create new Druid dataSources with the topic as its name.
+
+As an example, consider the following configuration (unrelated properties removed for clarity):
+
+```
+dataSources:
+  wikipedia:
+    spec:
+      dataSchema:
+        dataSource: wikipedia
+    properties:
+      topicPattern: wikipedia
+      topicPattern.priority: 2
+
+  wikipedia-lang:
+    spec:
+      dataSchema:
+        dataSource: wikipedia-lang
+    properties:
+      topicPattern: wikipedia.*
+      useTopicAsDataSource: true
+```
+
+Given the list of Kafka topics *[wikipedia, wikipedia-fr, wikipedia-es]*:
+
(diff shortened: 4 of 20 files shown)


gianm:
@dclim could you please address the merge conflicts?

build.sbt
```diff
@@ -19,6 +19,8 @@ val finagleVersion = "6.25.0"
 val twitterUtilVersion = "6.25.0"
 val samzaVersion = "0.8.0"
 val sparkVersion = "1.5.0"
+val kafkaVersion = "0.8.2.2"
```
gianm: Kafka 0.9.0.0, please

build.sbt
```diff
@@ -86,6 +88,19 @@ val sparkDependencies = Seq (
...
+  "org.apache.kafka" %% "kafka" % kafkaVersion % "optional"
+    exclude("org.slf4j", "slf4j-log4j12")
+    exclude("log4j", "log4j")
+    force(),
+  "io.airlift" % "airline" % airlineVersion % "optional",
+  "io.druid.extensions" % "druid-histogram" % druidVersion % "optional",
+  "org.apache.logging.log4j" % "log4j-api" % "2.3" % "optional",
+  "org.apache.logging.log4j" % "log4j-core" % "2.3" % "optional",
+  "org.apache.logging.log4j" % "log4j-slf4j-impl" % "2.3" % "optional",
+  "org.apache.logging.log4j" % "log4j-1.2-api" % "2.3" % "optional"
```
gianm: Please use logback to be consistent with the rest of tranquility (you can do `++ loggingDependencies` to get it)

build.sbt
```diff
@@ -86,6 +88,19 @@ val sparkDependencies = Seq (
     force()
 )
 
+val kafkaDependencies = Seq(
+  "org.apache.kafka" %% "kafka" % kafkaVersion % "optional"
+    exclude("org.slf4j", "slf4j-log4j12")
+    exclude("log4j", "log4j")
+    force(),
+  "io.airlift" % "airline" % airlineVersion % "optional",
+  "io.druid.extensions" % "druid-histogram" % druidVersion % "optional",
```
gianm: We can do this without including the histogram stuff explicitly.

kafka/src/main/java/com/metamx/tranquility/kafka/util/IngestionSpecHelper.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Read and validate ingestion spec files.
+ */
+public class IngestionSpecHelper
+{
+  private static final ObjectMapper mapper = new ObjectMapper();
```
gianm: This should be a `DruidGuicer.objectMapper` to make sure we load all extensions properly. That will also get the approximate histogram aggregator loaded if it's present, so the explicit dependency and code for that is not necessary.

kafka/src/main/java/com/metamx/tranquility/kafka/model/TranquilityConfig.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @JsonProperty
+  private boolean useTopicAsDataSource = false;
+
+  public int getBatchSize()
+  {
+    return batchSize;
+  }
+
+  public void setBatchSize(int batchSize)
```
gianm: If you want Jackson to use these validations rather than tweak the fields with reflection, I think you need to have the `@JsonProperty` here rather than on the field.

Although, if you don't need to be able to set things after construction, you could also use a `@JsonCreator` constructor and not have any setters. That is the standard way that we do immutable Jackson-enabled objects in java.

dclim: I tested it out and Jackson will actually use the setter if present, even if `@JsonProperty` is on the field and not the setter. Having said that, I like the suggestion of using a `@JsonCreator` constructor better.

dclim:
New changeset addresses the excellent feedback from @gianm and moves more of the validation into the model vs. a helper class. Thanks!

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,251 @@
...
+    startConsumers();
+  }
+
+  public void stop()
+  {
+    if (shutdown.compareAndSet(false, true)) {
+      log.info("Shutting down - attempting to flush buffers and commit final offsets");
+
+      try {
+        commitLock.writeLock().lockInterruptibly(); // prevent Kafka from consuming any more events
```
gianm: Generally you want to do the `lockInterruptibly` outside of the try block, because you only want to do the `unlock` if the lock is actually held.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,251 @@
...
+        writerController.stop(); // try to flush the remaining events to Druid
+        writerController.waitForAll(); // wait for all writers to finish flushing before committing offsets
+        consumerConnector.commitOffsets(); // update commit offset
+      } catch (InterruptedException e) {
+        Throwables.propagate(e);
```
gianm: This should also re-set the interrupted state on the thread with `Thread.currentThread().interrupt()`

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,251 @@
...
+      } finally {
+        consumerConnector.shutdown();
+        commitLock.writeLock().unlock();
```
gianm: Can `consumerConnector.shutdown()` fail? If so, the lock won't get unlocked.

Could `commitLock.writeLock().unlock()` be first so we don't have to worry about that, or is there a reason it's happening second here?

dclim: Agreed, it's good to move `commitLock.writeLock().unlock()` first and only call it if we've acquired the lock.

build.sbt
```diff
@@ -106,6 +108,14 @@ val serverDependencies = Seq(
   "org.eclipse.jetty" % "jetty-servlet" % jettyVersion
 ) ++ loggingDependencies
 
+val kafkaDependencies = Seq(
+  "org.apache.kafka" %% "kafka" % kafkaVersion % "optional"
```
gianm: Any reason why is this optional? Seems like it should be required.

build.sbt
```diff
@@ -106,6 +108,14 @@ val serverDependencies = Seq(
...
+    exclude("org.slf4j", "slf4j-log4j12")
+    exclude("log4j", "log4j")
+    force(),
+  "io.airlift" % "airline" % airlineVersion % "optional"
```
gianm: Any reason why is this optional? Seems like it should be required.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Spawns a number of threads to read messages from Kafka topics and write them by calling
+ * WriterController.getWriter(topic).send(). Will periodically call WriterController.flushAll() and when this completes
+ * will call ConsumerConnector.commitOffsets() to save the last written offset to ZooKeeper. This implementation
+ * guarantees that any events in Kafka will be read at least once even in case of a failure condition but does not
+ * guarantee that duplication will not occur.
+ */
+public class KafkaConsumer
+{
+  private static final Logger log = LoggerFactory.getLogger(KafkaConsumer.class);
```
gianm: Could you use the metamx common logger? That's more typical for druid project code.

```java
import com.metamx.common.logger.Logger;

private static final Logger log = new Logger(KafkaConsumer.class);
```

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,255 @@
...
+      final WriterController writerController
+  )
+  {
+    this.consumerConnector = getConsumerConnector(kafkaProperties);
+    this.topicFilter = new Whitelist(buildTopicFilter(schemaMap));
+    this.numThreads = numThreads;
+    this.commitMillis = commitMillis;
+    this.writerController = writerController;
+
+    this.consumerExec = Executors.newFixedThreadPool(numThreads,
```
gianm: (minor)

Bit more concise:

```java
Execs.multiThreaded(numThreads, "KafkaConsumer-%d")
```

dclim:
Thanks for the feedback @gianm , addressed comments and added unit tests.

build.sbt
```diff
@@ -140,6 +150,13 @@ val serverTestDependencies = Seq(
   "org.scalatra" %% "scalatra-test" % scalatraVersion % "test"
 )
 
+val kafkaTestDependencies = Seq(
+  "org.scalatest" %% "scalatest" % "2.2.5" % "test",
```
gianm: You shouldn't have to include scalatest, junit, or junit-interface here, since the kafka tests depend on the core tests and so they should get pulled in transitively (due to the `test->test` business in `dependsOn`)

kafka/src/main/java/com/metamx/tranquility/kafka/Main.java
```diff
@@ -0,0 +1,130 @@
...
+    final Properties props = PropertiesHelper.getApplicationArguments(propertiesFile);
+
+    final Properties kafkaProperties = new Properties();
+    kafkaProperties.setProperty("group.id", props.getProperty(KAFKA_GROUP_ID));
+    kafkaProperties.setProperty("zookeeper.connect",
+        String.format("%s%s", props.getProperty(ZK_CONNECT), props.getProperty(KAFKA_ZK_BASE_PATH)
+    ));
+    kafkaProperties.setProperty("zookeeper.session.timeout.ms", props.getProperty(ZK_SESSION_TIMEOUT_MS, "30000"));
+
+    final WriterController writerController = new WriterController(props.getProperty(ZK_CONNECT),
```
gianm: Apply java code style

kafka/src/main/java/com/metamx/tranquility/kafka/model/IngestionSpec.java
```diff
@@ -0,0 +1,60 @@
...
+  @JsonCreator
+  public IngestionSpec(
+      @JsonProperty("config") @Nullable TranquilityConfig config,
+      @JsonProperty("dataSchema") DataSchema dataSchema
+  )
+  {
+    this.config = (config != null ? config : new TranquilityConfig());
+    this.dataSchema = dataSchema;
+
+    Validate.notNull(dataSchema, "Spec must contain dataSchema attribute");
```
gianm: usually we use `Preconditions.checkNotNull` rather than the apache stuff

kafka/src/main/java/com/metamx/tranquility/kafka/model/IngestionSpec.java
```diff
@@ -0,0 +1,60 @@
...
+    Validate.notNull(dataSchema.getParser(), "Spec must contain parser attribute");
+    Validate.notNull(dataSchema.getParser().getParseSpec().getTimestampSpec(),
+                     "Spec must contain timestampSpec attribute");
+  }
+
+  public TranquilityConfig getConfig()
```
gianm: `@JsonProperty`

kafka/src/main/java/com/metamx/tranquility/kafka/model/IngestionSpec.java
```diff
@@ -0,0 +1,60 @@
...
+  {
+    return config;
+  }
+
+  public DataSchema getDataSchema()
```
gianm: `@JsonProperty`

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+
+    try {
+      Pattern.compile(topicPattern);
+    } catch (PatternSyntaxException e) {
+      throw new IAE(e, "[topicPattern] fields in %s must be valid regular expressions",
+                    PropertiesHelper.TRANQUILITY_SCHEMA_MAP
+      );
+    }
+  }
+
```
gianm: `@JsonProperty`

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+  public String getSpec()
+  {
+    return spec;
+  }
+
```
gianm: `@JsonProperty`

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Model used for mapping Kafka topics to Druid batch ingestion specs.
+ */
+public class TopicSchemaPair
+{
+  private String topicPattern;
```
gianm: These can be final

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+  @JsonCreator
+  public TopicSchemaPair(
+      @JsonProperty("topicPattern") String topicPattern,
+      @JsonProperty("spec") String spec
+  )
+  {
+    this.topicPattern = topicPattern;
+    this.spec = spec;
+
+    Validate.notEmpty(topicPattern, "[topicPattern] fields must be present in " +
```
gianm: `Preconditions.checkArgument(!topicPattern.isEmpty(), "…")`

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+  private String topicPattern;
+  private String spec;
+
+  @JsonCreator
+  public TopicSchemaPair(
+      @JsonProperty("topicPattern") String topicPattern,
+      @JsonProperty("spec") String spec
+  )
+  {
+    this.topicPattern = topicPattern;
```
gianm: `this.topicPattern = Preconditions.checkNotNull(topicPattern, "topicPattern");`

kafka/src/main/java/com/metamx/tranquility/kafka/model/TopicSchemaPair.java
```diff
@@ -0,0 +1,70 @@
...
+
+    Validate.notEmpty(topicPattern, "[topicPattern] fields must be present in " +
+                                    PropertiesHelper.TRANQUILITY_SCHEMA_MAP
+    );
+    Validate.notEmpty(spec, "[spec] fields must be present in " + PropertiesHelper.TRANQUILITY_SCHEMA_MAP);
+
+    try {
+      Pattern.compile(topicPattern);
+    } catch (PatternSyntaxException e) {
+      throw new IAE(e, "[topicPattern] fields in %s must be valid regular expressions",
```
gianm: Codestyle, I think (this looks a bit off)

kafka/src/main/java/com/metamx/tranquility/kafka/Main.java
```diff
@@ -0,0 +1,130 @@
...
+import io.airlift.airline.Command;
+import io.airlift.airline.Help;
+import io.airlift.airline.HelpOption;
+import io.airlift.airline.Option;
+import io.airlift.airline.SingleCommand;
+
+import javax.inject.Inject;
+import java.util.Properties;
+
+import static com.metamx.tranquility.kafka.util.PropertiesHelper.DRUID_CURATOR_PATH;
```
gianm: we generally avoid `import static`

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Read and validate application settings from a properties file.
+ */
+public class PropertiesHelper
```
gianm: I think it'd be good to use the same tranquility properties that `tranquility-server` does (https://github.com/druid-io/tranquility/blob/master/docs/server.md#configuration-reference). That stuff should probably be pulled out into core as it looks like more than one module will use it.

By "that stuff" I mean:

- `DataSourceConfig` in com.metamx.tranquility.server.config
- The code in com.metamx.tranquility.server.http.Main.createServlet that creates tranquilizers from a set of `DataSourceConfig`s

gianm: For the kafka properties, what do you think about taking anything that starts with `kafka.` and making it a kafka consumer property? That'd allow people to customize kafka consumer behaviors without us having to provide explicit support for them.

The only thing that we would need to make sure of is that we turn off autocommit, even if the user has set it on.

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
```diff
@@ -0,0 +1,149 @@
...
+    try (InputStream is = new FileInputStream(propertiesFile)) {
+      props.load(is);
+    } catch (IOException e) {
+      log.error("Could not read property file: %s, aborting.", propertiesFile);
+      Throwables.propagate(e);
+    }
+
+    validateProperties(props);
+
+    for (Map.Entry<?,?> entry : props.entrySet()) {
```
gianm: `props.stringPropertyNames` is a more type safe way of iterating properties.

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
```diff
@@ -0,0 +1,149 @@
...
+      String key = (String) entry.getKey();
+      if (key.startsWith("druid.extensions.")) {
+        System.setProperty(key, (String) entry.getValue());
+      }
+    }
+
+    IngestionSpecHelper.configPath = propertiesFile;
```
gianm: The fact that this is being set means that `IngestionSpecHelper` really wants to be a regular class (with a constructor that takes configPath) rather than a bunch of statics.

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
```diff
@@ -0,0 +1,149 @@
...
+
+    return defaultProps;
+  }
+
+  private static void validateProperties(Properties props)
+  {
+    try {
+      int threads = Integer.parseInt(props.getProperty(KAFKA_NUM_THREADS));
+      if (threads <= 0 || threads > 100) {
+        throw new IAE("%s must be between 1 and 100", KAFKA_NUM_THREADS);
```
gianm: what's wrong with more than 100 threads?

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
```diff
@@ -0,0 +1,149 @@
...
+      }
+    } catch (NumberFormatException e) {
+      throw new IAE(e, "Could not parse %s", KAFKA_NUM_THREADS);
+    }
+
+    try {
+      if (Integer.parseInt(props.getProperty(TRANQUILITY_COMMIT_MS)) < 5000) {
+        throw new IAE("%s must be greater than 5000", TRANQUILITY_COMMIT_MS);
```
gianm: IMO if people want to be crazy with commits, let them

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+  private final CuratorFramework curator;
+  private final String indexService;
+  private final String discoveryPath;
+  private final String specFile;
+  private final String topic;
+
+  private BeamPacketizer<Map<String, Object>> beamPacketizer;
+  private String timestampColumn;
+
+  public TranquilityEventWriter(String zkConnect, String indexService, String discoveryPath,
```
gianm: formatting looks off

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.Map;
+
+/**
+ * An EventWriter that pushes events to Druid through Tranquility.
+ */
+public class TranquilityEventWriter implements EventWriter
+{
+  private static final Logger log = new Logger(TranquilityEventWriter.class);
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final RetryPolicy retryPolicy = new ExponentialBackoffRetry(1000, 500, 30000);
```
gianm: generally, other than `log` we capitalize all the statics. So `MAPPER` and `RETRY_POLICY`

kafka/src/main/java/com/metamx/tranquility/kafka/writer/WriterController.java
```diff
@@ -0,0 +1,107 @@
...
+  private String discoveryPath;
+  private List<TopicSchemaPair> topicSchemaPairs;
+  private Map<String, SynchronizedEventWriter> writers = new HashMap<>();
+
+  public WriterController(String zkConnect, String indexService, String discoveryPath, String schemaMap) {
+    this.zkConnect = zkConnect;
+    this.indexService = indexService;
+    this.discoveryPath = discoveryPath;
+
+    try {
```
gianm: It'd be nicer to pass in the actual list of TopicSchemaPairs rather than a string that gets deserialized. Generally we try to give classes like this what they actually need, and handle the instantiation of it somewhere else (either the Main class or a factory or something like that)

kafka/src/main/java/com/metamx/tranquility/kafka/writer/WriterController.java
```diff
@@ -0,0 +1,107 @@
...
+      Throwables.propagate(e);
+    }
+
+    log.info("Ready: [topicPattern] -> spec mappings (ordered):");
+    for (TopicSchemaPair pair : topicSchemaPairs) {
+      log.info("  [%s] -> %s", pair.getTopicPattern(), pair.getSpec());
+    }
+  }
+
+  public EventWriter getWriter(String topic) {
```
gianm: This method appears to be accessed concurrently, but it is not thread safe.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/WriterController.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+public class WriterController
+{
+  private static final Logger log = new Logger(WriterController.class);
+
+  private String zkConnect;
+  private String indexService;
+  private String discoveryPath;
+  private List<TopicSchemaPair> topicSchemaPairs;
+  private Map<String, SynchronizedEventWriter> writers = new HashMap<>();
```
gianm: Access to `writers` should either be synchronized or it should be a ConcurrentHashMap.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/SynchronizedEventWriter.java
```diff
@@ -0,0 +1,121 @@
...
+      if (pendingEvents == 0) {
+        flushAwaitLock.notifyAll();
+      }
+    }
+  }
+
+  @Override
+  public void init(EventWriterListener listener)
+  {
+    // do nothing
```
gianm: This should throw an exception as it appears intended to never actually be called. (the listener would be ignored)

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+  {
+    final IngestionSpec ingestionSpec = IngestionSpecHelper.parseSpecFile(specFile);
+    final DataSchema schema = ingestionSpec.getDataSchema();
+    final TranquilityConfig tranquilityConfig = ingestionSpec.getConfig();
+    final String dataSource = tranquilityConfig.useTopicAsDataSource() ? topic : schema.getDataSource();
+    final String firehosePattern = "tranquility-kafka:" + indexService + ":" + dataSource + ":%s";
+
+    timestampColumn = schema.getParser().getParseSpec().getTimestampSpec().getTimestampColumn();
+
+    final Timestamper<Map<String, Object>> timestamper = new Timestamper<Map<String, Object>>()
```
gianm: This Timestamper won't work for timestamp formats other than ISO or millis. If you use the DruidBeams builder that takes a DruidEnvironment and FireDepartment it makes a more flexible Timestamper.

gianm: This will probably happen on its own if you end up pulling out the configuring code from server's `Main` into core, as it does do this.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+      return false;
+    }
+
+    beamPacketizer.send(map);
+
+    return true;
+  }
+
+  @Override
+  public void flush() throws InterruptedException
```
gianm: This needs to be synchronized, `BeamPacketizer`s are not thread safe.

gianm: Or you could switch to a `Tranquilizer` since they are thread safe.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+  {
+    if (beamPacketizer == null) {
+      throw new IllegalStateException("Call init() first");
+    }
+
+    Map<String, Object> map;
+    try {
+      map = mapper.readValue(message, new TypeReference<HashMap<String, Object>>() { });
+    } catch (Exception e) {
+      log.error(e, "MESSAGE REJECTED - Failed to parse input data as JSON for: %s", message);
```
gianm: It's usually better to make stuff like this a `debug` and increment a counter that we report periodically, to avoid log spam if there are a flurry of bad messages.

gianm: Or report at most one message every minute (or whatever) and also report a counter periodically. Generally anything that avoids potentially huge amounts of log messages.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+      return false;
+    }
+
+    try {
+      new DateTime(map.get(timestampColumn));
+    } catch (Exception e) {
+      log.error(e, "MESSAGE REJECTED - Failed to parse timestamp for: %s", map.get(timestampColumn));
```
gianm: similar comment to above; also, this should use an appropriate Timestamper in case the time format isn't ISO or millis.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,231 @@
...
+  {
+    try {
+      beamPacketizer.close();
+    } catch (IllegalStateException e) {
+      log.info(e, "Exception while stopping Tranquility");
+    }
+    curator.close();
+  }
+
+  private List<String> getDimensions(DataSchema schema) {
```
gianm: I think these won't be needed if you use the DruidBeams FireDepartment builder

kafka/src/test/java/com/metamx/tranquility/kafka/KafkaConsumerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    EasyMock.expect(mockWriterController.getWriter(topic)).andReturn(mockEventWriter).times(numMessages);
+    mockWriterController.stop();
+    mockWriterController.waitForAll();
+    EasyMock.replay(mockWriterController, mockEventWriter);
+
+    // commitMillis is set high enough that the commit thread should not run during the test
+    KafkaConsumer kafkaConsumer = new KafkaConsumer(consumerProperties, schemaMap, 2, 999999, mockWriterController);
+    kafkaConsumer.start();
+
+    Thread.sleep(500); // ugly, but helps give Kafka time to settle before sending messages
```
gianm: Why is this needed? The consumer should be able to retry until Kafka is up, and this seems like the kind of thing that will cause random test failures.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,255 @@
...
+                // is returned by next(). In order to guarantee at-least-once message delivery, we need to a) set
+                // autocommit enable to false so the consumer will not automatically commit offsets to Zookeeper, and
+                // b) synchronize calls of kafkaIterator.next() with the commit thread so that we don't read messages
+                // and then call consumerConnector.commitOffsets() before those messages have been flushed through
+                // Tranquility into the indexing service.
+                commitLock.readLock().lockInterruptibly();
+
+                try {
+                  MessageAndMetadata<byte[], byte[]> data = kafkaIterator.next();
+                  writerController.getWriter(data.topic()).send(new String(data.message(), Charsets.UTF_8));
```
gianm: It should be possible to do this without converting to a String. Jackson can parse the `byte[]` directly.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,255 @@
...
+      writerController.waitForAll(); // wait for all writers to finish flushing before committing offsets
+
+      final long commitStartTime = System.currentTimeMillis();
+      consumerConnector.commitOffsets();
+
+      final long finishedTime = System.currentTimeMillis();
+      log.info(
+          "Flushed %s pending messages in %sms and committed offsets in %sms.",
+          eventsFlushed.isEmpty() ? "0" : eventsFlushed,
+          finishedTime - flushStartTime,
```
gianm: I think this should be `commitStartTime - flushStartTime`

kafka/src/main/java/com/metamx/tranquility/kafka/writer/SynchronizedEventWriter.java
gianm: It may also make sense for SynchronizedEventWriter to not actually be an EventWriter.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/SynchronizedEventWriter.java
```diff
@@ -0,0 +1,121 @@
...
+          public void onSuccess(Object o)
+          {
+            decrementPending();
+          }
+
+          @Override
+          public void onFailure(Throwable throwable, Object o)
+          {
+            decrementPending();
+            log.error(throwable, "Failed to send event: %s", (o == null ? "null" : o.toString()));
```
gianm: It's usually better to make stuff like this a debug and increment a counter that we report periodically, to avoid log spam if there are a flurry of bad messages. Or report at most one message every minute (or whatever) and also report a counter periodically. Generally anything that avoids potentially huge amounts of log messages.

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
dclim: @gianm fair enough :P

kafka/src/main/java/com/metamx/tranquility/kafka/util/PropertiesHelper.java
dclim: @gianm great idea

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
dclim: @gianm flush() only ever gets called by a single commit thread, it's send() that gets called from multiple places and is thus synchronized. Having said that I'll be re-working this to use Tranquilizer.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/WriterController.java
dclim: @gianm :+1:

kafka/src/test/java/com/metamx/tranquility/kafka/KafkaConsumerTest.java
dclim: @gianm yeah, was definitely trying to avoid magic sleeps and brittle tests. I can test it some more without the sleep and see if it fails, but I was seeing occasional sporadic failures before because some messages were getting lost

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
gianm: *All* access to a non-thread-safe object needs to be synchronized somehow, since without synchronization points there is no guarantee that different threads will see consistent versions of those data structures ("visibility" on http://gee.cs.oswego.edu/dl/cpj/jmm.html)

core/src/main/scala/com/metamx/tranquility/config/ConfigHelper.scala
```diff
@@ -0,0 +1,114 @@
+package com.metamx.tranquility.config
+
```
gianm: Optimize imports / apply codestyle (this file looks like it is a little bit off on each of those)

core/src/main/scala/com/metamx/tranquility/config/ConfigHelper.scala
```diff
@@ -0,0 +1,114 @@
...
+    ): (T, Map[String, DataSourceConfig[T]], Dict) =
+  {
+    val yaml = in.withFinally(_.close()) { in =>
+      dict(Yaml.load(in))
+    }
+    val globalProperties: Dict = getAsDict(yaml, "properties")
+    val globalConfig = mapConfig(globalProperties, clazz)
+
+    for (prop <- globalProperties.filterKeys { _ startsWith("druid.extensions.") }) {
+      sys.props += (prop._1 -> prop._2.toString)
```
gianm: Hmm, any editing of system properties is kind of gnarly. It's probably ok since it's meant to be called by a main, but would suggest two things:

- Comment on `readConfigYaml` indicating that it may modify system properties
- Only set the system property if it's not already set. This is because we usually want to respect a property from the commandline over one supplied in a config file.

gianm: Also a comment here indicating why we need to modify system properties (presumably because we need those set in order for the DruidGuicer to pick up extensions)

dclim: @gianm Good idea, changes made.

core/src/main/scala/com/metamx/tranquility/config/ConfigHelper.scala
```diff
@@ -0,0 +1,125 @@
...
+    clazz: Class[T]
+    ): (T, Map[String, DataSourceConfig[T]], Dict) =
+  {
+    val yaml = in.withFinally(_.close()) { in =>
+      dict(Yaml.load(in))
+    }
+    val globalProperties: Dict = getAsDict(yaml, "properties")
+    val globalConfig = mapConfig(globalProperties, clazz)
+
+    for (prop <- globalProperties.filterKeys {
```
gianm: generally prefer to format this on one line as `globalProperties.filterKeys(_.startsWith("druid.extensions."))`

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
```diff
@@ -186,9 +119,9 @@ object ServerMain extends App with Logging
 
   def createJettyServer(
     lifecycle: Lifecycle,
-    globalConfig: GlobalConfig,
+    globalConfig: ServerConfig,
     servlet: TranquilityServlet
-  ): Server =
+    ): Server =
```
gianm: I'm not sure why your formatter is adding these two extra spaces on all the multiline method arg lists. Does applying the codestyle remove the spaces? (cmd-alt-L).

If not I'm ok letting it slide :P

server/src/test/scala/com/metamx/tranquility/server/ServerConfigTest.scala
```diff
@@ -44,6 +44,5 @@ class ServerMainTest extends FunSuite with ShouldMatchers
     fooConfig.fireDepartment.getTuningConfig.getWindowPeriod should be(new Period("PT30S"))
     fooConfig.config.zookeeperConnect should be("zk.example.com")
     fooConfig.config.taskPartitions should be(3)
-    fooConfig.config.druidBeamConfig.firehoseGracePeriod should be(new Period("PT1S"))
```
gianm: Why is this gone? This property should still work.

dclim: That's a very good question, I didn't explicitly remove it. Something weird with the merge, I'll look into it.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,273 @@
...
+    try {
+      final long flushStartTime = System.currentTimeMillis();
+      final Map<String, MessageCounters> messageCounters = writerController.flushAll(); // blocks until complete
+
+      final long commitStartTime = System.currentTimeMillis();
+      consumerConnector.commitOffsets();
+
+      final long finishedTime = System.currentTimeMillis();
+      Map<String, MessageCounters> countsSinceLastCommit = new HashMap();
+      for (Map.Entry<String, MessageCounters> entry : messageCounters.entrySet()) {
```
gianm: At some point we should add some metrics collection ability but that does not have to be done in this PR

gianm: (For Core and Server too)

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,273 @@
...
+        try {
+          while (!Thread.currentThread().isInterrupted()) {
+            Thread.sleep(Math.max(commitMillis - (System.currentTimeMillis() - lastFlushTime), 0));
+            commit();
+            lastFlushTime = System.currentTimeMillis();
+          }
+        }
+        catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+          log.info(e, "Commit thread interrupted.");
```
gianm: The `e, ` here is going to make the log message scarier than it needs to be, & can probably be skipped.

kafka/src/main/java/com/metamx/tranquility/kafka/KafkaConsumer.java
```diff
@@ -0,0 +1,273 @@
...
+                    MessageAndMetadata<byte[], byte[]> data = kafkaIterator.next();
+                    writerController.getWriter(data.topic()).send(data.message());
+                  }
+                  finally {
+                    commitLock.readLock().unlock();
+                  }
+                }
+              }
+              catch (InterruptedException e) {
+                log.info(e, "Consumer thread interrupted.");
```
gianm: Similar comment about the `e, ` (interrupting is going to happen on normal shutdowns)

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
dclim: Hmm no, applying the formatter adds those spaces back when I remove them.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/WriterController.java
```diff
@@ -0,0 +1,113 @@
...
+        {
+          @Override
+          public int compare(DataSourceConfig<TranquilityKafkaConfig> o1, DataSourceConfig<TranquilityKafkaConfig> o2)
+          {
+            return o2.config().getTopicPatternPriority().compareTo(o1.config().getTopicPatternPriority());
+          }
+        }
+    );
+
+    log.info("Ready: [topicPattern] -> dataSource mappings (priority):");
```
gianm: the `(priority)` is probably not needed as it's explicitly mentioned in the following log messages.

kafka/src/main/java/com/metamx/tranquility/kafka/Main.java
```diff
@@ -0,0 +1,149 @@
...
+    final Properties kafkaProperties = new Properties();
+    for (Map.Entry<String, Object> entry : globalProperties.entrySet()) {
+      String key = entry.getKey();
+      log.info("%s = %s", key, globalProperties.get(key).toString());
+      if (key.startsWith("kafka.")) {
+        kafkaProperties.setProperty(key.replaceFirst("kafka.", ""), globalProperties.get(key).toString());
+      }
+    }
+
+    // set the critical Kafka configs again from TranquilityKafkaConfig so it picks up the defaults
```
gianm: This should probably throw an error if a property is specified in both places but have different values, since otherwise people with mismatched configs may have a hard time debugging.

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
gianm: That's a bummer but ok let's not worry about it

kafka/src/main/java/com/metamx/tranquility/kafka/Main.java
dclim: For the first two (kafka.group.id and kafka.zookeeper.connect) it's not possible for the user to specify a different value in both places since they're both based on the same property. For the last one, it could be possible so I can add a check for that one specifically.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,161 @@
...
+    this.dataSourceConfig = dataSourceConfig;
+
+    final int zkTimeout = Ints.checkedCast(
+        dataSourceConfig.config()
+                        .zookeeperTimeout()
+                        .toStandardDuration()
+                        .getMillis()
+    );
+
+    curator = CuratorFrameworkFactory.builder()
```
gianm: Curator and FinagleRegistry instances should be shared amongst all TranquilityEventWriters, as each one maintains threads and connections-- see how ServerMain handles that for an example.

gianm: That'll allow scaling to more data sources without overloading things.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,161 @@
...
+        log.debug(e, "%d message(s) failed to parse as JSON and were rejected", rejectedLogCount);
+      }
+      if (dataSourceConfig.config().reportDropsAsExceptions()) {
+        Throwables.propagate(e);
+      }
+
+      return;
+    }
+
+    simpleTranquilizerAdapter.send(
```
gianm: This is gonna potentially throw an exception caused by a totally unrelated message (due to how STA works) but I think that's okay, since the consumer is set up to stop all threads and exit if any send throws an exception. A comment pointing out this situation would be good.

kafka/src/main/java/com/metamx/tranquility/kafka/model/TranquilityKafkaConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import scala.collection.JavaConversions;
+
+/**
+ * Configuration object which extends Tranquility configuration with Kafka specific parameters.
+ */
+public abstract class TranquilityKafkaConfig extends TranquilityConfig
+{
+  public TranquilityKafkaConfig()
+  {
+    GlobalProperties_$eq(
```
gianm: i wonder what _$eq is

gianm: This might work better with Java if TranquilityConfig were an abstract class and the constructor took what the global properties should be.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    final Tranquilizer<scala.collection.immutable.Map<String, Object>> tranquilizer =
+        ConfigHelper.createTranquilizerWithLocation(
+            dataSourceConfig,
+            finagleRegistry,
+            curator,
+            DruidLocation.create(
+                dataSourceConfig.config().druidIndexingServiceName(),
+                dataSourceConfig.config().useTopicAsDataSource()
```
gianm: hmm, this useTopicAsDataSource feature is kind of cool, but also strange in that it means the `dataSource` in the spec will be ignored. I think that deserves being called out in the docs

gianm:
@dclim looks good other than remaining comments. Could you also add some docs along the lines of what exists in `docs/server.md`? (and a reference to them on the main `README.md`)

kafka/src/main/java/com/metamx/tranquility/kafka/model/TranquilityKafkaConfig.java
dclim: _$eq is the internal setter created by Scala. Giving TranquilityConfig a constructor sounds like a good option.

kafka/src/main/java/com/metamx/tranquility/kafka/writer/TranquilityEventWriter.java
dclim: Definitely will call it out.

core/src/main/scala/com/metamx/tranquility/config/config.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.joda.time.Period
+import org.skife.config.Config
+
+abstract class TranquilityConfig(globalProperties: Set[String]) extends DiscoConfig
+{
+  def this() {
+    this(Set[String]())
+  }
+
+  var GlobalProperties = Set[String]() ++ globalProperties
```
gianm: this can/should be a `val`

docs/kafka.md
```diff
@@ -0,0 +1,145 @@
...
+## Configuration reference
+
+With the exception of the global properties (which must be provided at the top level), most properties can be provided
+at the top level or per-dataSource. If the same property is provided at both levels, the per-dataSource version will
+take precedence. This provides a way to set default properties that apply to all dataSources, while still allowing
+customization for certain dataSources.
+
+### Global properties
+
+|Property|Description|Default|
```
gianm: Anything that starts in `kafka.` is passed to the consumer, right? that should be mentioned here too.

gianm:
:+1: after doc changes

gianm:
@xvrl @drcrallen @nishantmonu51 @fjy anyone able to do another review of this?

dclim:
Changes made and commits squashed. Thanks Gian.

fjy:
@gianm @dclim There's license headers missing on a lot of files. Other than that Im :+1:

dclim:
Will add missing license headers.

dclim:
@gianm @fjy Headers are in

fjy:
:+1:

PR #67, opened 2015-12-14
Issue#53

rohitkochar:
Fixing Issue#53.
Used task duration of realtime tasks to form the service name.


rohitkochar:
Found an issue fix..hence closing this pull req,will raise another later

PR #72, opened 2015-12-20
Issue#53

rohitkochar:
Fixes issue 53 by considering the duration of RT task while forming the "BaseFireHoseId" which is also used to form the name of Finangle Service.

Thought Process:
1. If Segment Granularity is less than Task Duration 
Lets say Segment Granularity is "Minute" and Task Duration is "2 hours" than as per this fix "BaseFireHoseId"  will be "datasource-hour of day-minute of hour-partition" so it includes both hour part and minute part of the time hence multiple task belong to same minute of different hour will have different "BaseFireHoseId" hence no clash.

2.  If Segment Granularity is more than Task Duration 
Lets say Segment Granularity is "Hour" and Task Duration is "5 minutes" than as per this fix "BaseFireHoseId"  will be "datasource-minute of hour-partition" so it includes only "minute" part of the time but since task duration is only 5 minutes,by the time new task is created(which will happen once per hour") no previous task of the same minute of previous hours would exist hence no collision. 

Hence segmenet granularity is redundant info while forming "BaseFirehoseId" and only task duration needs to be considered.


core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -216,26 +219,23 @@ class DruidBeamMaker[A: Timestamper](
...
+
+    val cycleBucket = taskDuration match {
+      case i if(i.toStandardSeconds.getSeconds < DateTimeConstants.SECONDS_PER_HOUR) => "%02d".format(tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds().getSeconds < DateTimeConstants.SECONDS_PER_DAY) => "%02d-%02d" .format(tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds().getSeconds < DateTimeConstants.SECONDS_PER_DAY *28) => "%02d-%02d-%02d" .format(tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds.getSeconds < DateTimeConstants.SECONDS_PER_DAY*365) => "%02d-%02d-%02d-%02d" .format(tsUtc.monthOfYear().get,tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds.getSeconds > DateTimeConstants.SECONDS_PER_DAY*365) => "%02d-%02d-%02d-%02d-%02d" .format(tsUtc.yearOfCentury().get,tsUtc.monthOfYear().get,tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case x => throw new IllegalArgumentException("No gross firehose id hack for task duration [%s]" format x)
```
gianm: It looks like you've actually covered all the cases, so you can remove this exception.

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -216,26 +219,23 @@ class DruidBeamMaker[A: Timestamper](
...
-      case Granularity.YEAR => tsUtc.yearOfCentury().get
-      case x => throw new IllegalArgumentException("No gross firehose id hack for granularity[%s]" format x)
+
+    val cycleBucket = taskDuration match {
+      case i if(i.toStandardSeconds.getSeconds < DateTimeConstants.SECONDS_PER_HOUR) => "%02d".format(tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds().getSeconds < DateTimeConstants.SECONDS_PER_DAY) => "%02d-%02d" .format(tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds().getSeconds < DateTimeConstants.SECONDS_PER_DAY *28) => "%02d-%02d-%02d" .format(tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds.getSeconds < DateTimeConstants.SECONDS_PER_DAY*365) => "%02d-%02d-%02d-%02d" .format(tsUtc.monthOfYear().get,tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
+      case i if(i.toStandardSeconds.getSeconds > DateTimeConstants.SECONDS_PER_DAY*365) => "%02d-%02d-%02d-%02d-%02d" .format(tsUtc.yearOfCentury().get,tsUtc.monthOfYear().get,tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
```
gianm: This looks like it's supposed to be a fallback, so it could just be:

```scala
case i => "%02d-%02d-%02d-%02d-%02d" .format(tsUtc.yearOfCentury().get,tsUtc.monthOfYear().get,tsUtc.dayOfMonth().get,tsUtc.hourOfDay().get,tsUtc.minuteOfHour().get)
```

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -216,26 +219,23 @@ class DruidBeamMaker[A: Timestamper](
 
 object DruidBeamMaker
 {
-  def generateBaseFirehoseId(dataSource: String, segmentGranularity: Granularity, ts: DateTime, partition: Int) = {
+  def generateBaseFirehoseId(dataSource: String,taskDuration:Period, ts: DateTime, partition: Int) = {
     // Not only is this a nasty hack, it also only works if the RT task hands things off in a timely manner. We'd rather
     // use UUIDs, but this creates a ton of clutter in service discovery.
+    //It may also break if task duration is changed across RT tasks
```
gianm: In what way will this break if task duration is changed?

gianm:
@rohitkochar thanks for the contribution!

I made a couple of comments on the diff; could you also please fix the conflicts and apply the Scala code-style at: https://github.com/druid-io/druid/raw/master/intellij_formatting.jar

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
rohitkochar: Lets say earlier duration is less than an hour than the firehose id will be "datasource-<minute of hour>-partition" and later if application is restarted and task duration is changed to more than an hour than tranquility won't be able to identify the older running task as it would be looking for "datasource-<hour ofday>-<minute of hour>-partition"

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
rohitkochar: taken care

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
rohitkochar: taken care

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
gianm: ah, I see. This should actually be okay, since tranquility remembers the firehoseId when it creates a task, rather than re-generating it each time. It's stored in the beam metadata in zk and can be restored after app restart.

gianm: So if that's the only concern I think it's ok to remove that "it may break" comment.

rohitkochar: taken care

rohitkochar:
@gianm I have resolved the merge conflicts and have also applied the code-style.
Please review and merge.

rohitkochar:
@gianm 
Can you please review this?

gianm:
Hey @rohitkochar, really sorry for the delay on this one! #171 contains a similar patch that I think should work for you too (segment granularity minute, duration 2 hour). If you have a chance to try it out please let me know.

FWIW the real fix for this issue is setting `druidBeam.taskLocator = overlord`, which will work starting in Druid 0.9.1. At that point this cycleBucket stuff won't be needed.

gianm:
I believe this to be addressed in the short term by #171, and in the slightly less short term by `druidBeam.taskLocator = overlord` in druid 0.9.1. so, closing.

PR #73, opened 2015-12-21
Update to Druid 0.8.2, and switch to JSON API instead of directly using Task classes.

### LINKED ISSUE #47: DataSchema constructor mismatch with druid 0.8.2
just recording that I'm seeing this error when running tranquility with druid 0.9 jars:
  2015-11-04T23:00:50,557 ERROR [ClusteredBeam-ZkFuturePool-5998d9e6-6fc8-45c4-9769-163ed08247f7] com.metamx.tranquility.beam.ClusteredBeam - Failed to update cluster state: overlord/tranquility_test
java.lang.NoSuchMethodError: io.druid.segment.indexing.DataSchema.<init>(Ljava/lang/String;Lio/druid/data/input/impl/InputRowParser;[Lio/druid/query/aggregation/AggregatorFactory;Lio/druid/segment/indexing/granularity/GranularitySpec;)V
        at com.metamx.tranquility.druid.DruidBeamMaker.com$metamx$tranquility$druid$DruidBeamMaker$$taskObject(DruidBeamMaker.scala:84) ~[tranquility-core_2.11-0.6.2.jar:0.6.2]

This is doing "new DataSchema" with 4 arguments, but the DataSchema constructor has 5 arguments as of druid PR #1695, which I think is in druid 0.8.2.

gianm:
Fixes #47.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
@@ -18,6 +18,7 @@
  */
 package com.metamx.tranquility.druid
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.metamx.common.Granularity
 import com.metamx.common.scala.Logging
 import com.metamx.common.scala.untyped._
@@ -29,27 +30,13 @@ import com.metamx.tranquility.typeclass.ObjectWriter
 import com.metamx.tranquility.typeclass.Timestamper
 import com.twitter.util.Await
 import com.twitter.util.Future
-import io.druid.data.input.impl.JSONParseSpec
-import io.druid.data.input.impl.MapInputRowParser
 import io.druid.data.input.impl.TimestampSpec
-import io.druid.indexing.common.task.RealtimeIndexTask
-import io.druid.indexing.common.task.Task
-import io.druid.indexing.common.task.TaskResource
-import io.druid.segment.indexing.granularity.UniformGranularitySpec
-import io.druid.segment.indexing.DataSchema
-import io.druid.segment.indexing.RealtimeIOConfig
-import io.druid.segment.indexing.RealtimeTuningConfig
-import io.druid.segment.realtime.FireDepartment
-import io.druid.segment.realtime.firehose.ClippedFirehoseFactory
-import io.druid.segment.realtime.firehose.EventReceiverFirehoseFactory
-import io.druid.segment.realtime.firehose.TimedShutoffFirehoseFactory
-import io.druid.segment.realtime.plumber.NoopRejectionPolicyFactory
-import io.druid.segment.realtime.plumber.ServerTimeRejectionPolicyFactory
-import io.druid.timeline.partition.LinearShardSpec
-import org.joda.time.chrono.ISOChronology
+import java.{util => ju}
 import org.joda.time.DateTime
 import org.joda.time.Interval
+import org.joda.time.chrono.ISOChronology
 import org.scala_tools.time.Implicits._
+import scala.collection.JavaConverters._
 import scala.util.Random
 
 class DruidBeamMaker[A: Timestamper](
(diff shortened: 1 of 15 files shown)


core/src/main/scala/com/metamx/tranquility/druid/DruidRollup.scala
```diff
@@ -44,15 +44,10 @@ class DruidRollup(
...
-    val spatialDimensionNames = dimensions.spec.getSpatialDimensions.asScala.map(_.getDimName)
+    val dimensionNames = dimensions.knownDimensions
     val metricNames = aggregators.map(_.getName)
 
-    val allColumnNames = Seq(DruidRollup.InternalTimeColumnName) ++
-      dimensionNames ++
-      spatialDimensionNames ++
-      metricNames
-
+    val allColumnNames = Seq(DruidRollup.InternalTimeColumnName) ++ dimensionNames ++ metricNames
```
drcrallen: Dropping spatial dimensions?

drcrallen: oh I see, on down

core/src/test/resources/druid-broker.properties
```diff
@@ -1,23 +1,30 @@
...
 druid.host=localhost
 druid.port=:DRUIDPORT:
 druid.service=druid:tranquility:broker
+
 druid.zk.service.host=:ZKCONNECT:
+
 druid.processing.numThreads=1
 druid.processing.buffer.sizeBytes=8388608
+
+druid.selectors.indexing.serviceName=druid:tranquility:indexer
```
drcrallen: are changes in this file part of this patch?

gianm: I needed to add a coordinator since 0.8.2 overlords are more finicky about wanting to create their DB tables, and the coordinator by default is the derby metadata server (in 0.8.1 it failed & continued).

Along with adding a coordinator I took the opportunity to update all the runtime properties files.

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -82,59 +70,79 @@ class DruidBeamMaker[A: Timestamper](
...
+            "type" -> "receiver",
+            "serviceName" -> location.environment.firehoseServicePattern.format(firehoseId),
+            "bufferSize" -> config.firehoseBufferSize
+          ).asJava
+        ).asJava
+      ).asJava
+    ).asJava
+    val tuningConfigMap = Map(
+      "type" -> "realtime",
+      "maxRowsInMemory" -> druidTuning.maxRowsInMemory,
```
drcrallen: deprecated

see https://github.com/druid-io/druid/pull/1922

gianm: I thought `maxRowsInMemory` was the good one and `rowFlushBoundary` was the deprecated one.

drcrallen: I think you are correct. My brain is backwards today

drcrallen:
Given the delicate nature of string parameters (they can change and silently be ignored) is it possible to add some unit tests that ensure the objects generated are those expected?

I just worry that down the line names will change and not be noticed.

gianm:
@drcrallen are you imagining a test that is something like, the schema json generated by DruidBeamMaker with some set of parameters should be some particular map?

drcrallen:
@gianm would it be possible to generate the json with the DruidBeamMaker, then parse the task with a druid ObjectMapper and compare the resultant values to the expected set values?

gianm:
@drcrallen yes, that's possible. It will be brittle to Druid updates but that's probably okay if it's limited to the tests.

drcrallen:
@gianm my thoughts exactly. it should be able to be written so that it does not call constructors. And the getters tend not to change unless the variable names change also (in which case it would need to be double-checked anyways)

fjy:
:+1:

gianm:
@drcrallen added a serde test

drcrallen:
Thanks @gianm I appreciate the awesome tests

rasahner:
My thanks, too, for addressing this.

PR #75, opened 2015-12-22
Add Beam.sendBatch (returns a BitSet of successes)

### LINKED ISSUE #56: Modify Beam trait so callers can tell which events were dropped

gianm:
Add Beam.sendBatch (returns a BitSet of successes), fixes #56.

Also:

- Deprecate Beam.propagate
- Make Tranquilizer's MessageDroppedException a singleton
- Improve ClusteredBeam tests and add tests involving dropping events

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
+++ b/core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
@@ -19,26 +19,69 @@
 package com.metamx.tranquility.beam
 
 import com.twitter.util.Future
+import scala.collection.immutable.BitSet
+import scala.collection.mutable
 
 /**
- * Beams can accept events and forward them along. The propagate method may throw a DefunctBeamException, which means
- * the beam should be discarded (after calling close()).
- */
-trait Beam[A]
+  * Beams can accept messages and forward them along. The propagate method may throw a DefunctBeamException, which
+  * means the beam should be discarded (after calling close()).
+  */
+trait Beam[MessageType]
 {
   /**
-   * Request propagation of events. The operation may fail in various ways, which tend to be specific to
-   * the implementing class. Returns the number of events propagated.
-   */
-  def propagate(events: Seq[A]): Future[Int]
+    * Request propagation of messages. The return value indicates the number of messages known to be sent
+    * successfully. Note that for some implementations, it is possible for a message to be sent and for the ack to
+    * be lost (e.g. if the send is occurring over a network).
+    * @param messages a batch of messages
+    * @return the number of messages propagated
+    */
+  @deprecated("use sendBatch", "0.7.0")
+  final def propagate(messages: Seq[MessageType]): Future[Int] = {
+    sendBatch(messages).map(_.size)
+  }
 
   /**
-   * Signal that no more events will be sent. Many implementations need this to do proper cleanup.
-   */
+    * Request propagation of messages. The returned bitset contains the indexes of messages known to be sent
+    * successfully. Note that for some implementations, it is possible for a message to be sent and for the ack to be
+    * lost (e.g. if the send is occurring over a network).
+    * @param messages a batch of messages
+    * @return a bitset containing indexes of messages that were sent successfully
+    */
+  def sendBatch(messages: Seq[MessageType]): Future[BitSet]
+
+  /**
+    * Signal that no more messages will be sent. Many implementations need this to do proper cleanup. This operation
+    * may happen asynchronously.
+    * @return future that resolves when closing is complete
+    */
   def close(): Future[Unit]
 }
 
 class DefunctBeamException(s: String, t: Throwable) extends Exception(s, t)
 {
   def this(s: String) = this(s, null)
+}
+
+object Beam
+{
+  /**
+    * Beam.index(xs) is like xs.zipWithIndex, but always returns an IndexedSeq so lookups by index will be efficient.
+    * @param xs elements
+    * @return indexed seq equivalent to xs.zipWithIndex
+    */
+  def index[A](xs: Seq[A]): IndexedSeq[(A, Int)] = {
+    val indexed = Vector.newBuilder[(A, Int)]
+    var i = 0
+    for (message <- xs) {
+      indexed += ((message, i))
+      i += 1
+    }
+    indexed.result()
+  }
+
+  def mergeBitsets[A](bitsets: Iterable[BitSet]): BitSet = {
+    val merged = mutable.BitSet()
+    bitsets.foreach(merged ++= _)
+    merged.toImmutable
+  }
 }
--- a/core/src/main/scala/com/metamx/tranquility/beam/ClusteredBeam.scala
+++ b/core/src/main/scala/com/metamx/tranquility/beam/ClusteredBeam.scala
@@ -43,39 +43,40 @@ import org.joda.time.Interval
 import org.joda.time.chrono.ISOChronology
 import org.scala_tools.time.Implicits._
 import scala.collection.JavaConverters._
+import scala.collection.immutable.BitSet
 import scala.collection.mutable
 import scala.language.reflectiveCalls
 import scala.util.Random
 
 /**
- * Beam composed of a stack of smaller beams. The smaller beams are split across two axes: timestamp (time shard
- * of the data) and partition (shard of the data within one time interval). The stack of beams for a particular
- * timestamp are created in a coordinated fashion, such that all ClusteredBeams for the same identifier will have
- * semantically identical stacks. This interaction is mediated through zookeeper. Beam information persists across
- * ClusteredBeam restarts.
- *
- * In the case of Druid, each merged beam corresponds to one segment partition number, and each inner beam corresponds
- * to either one index task or a set of redundant index tasks.
- *
- * {{{
- *                                            ClusteredBeam
- *
- *                                   +-------------+---------------+
- *               2010-01-02T03:00:00 |                             |   2010-01-02T04:00:00
- *                                   |                             |
- *                                   v                             v
- *
- *                         +----+ Merged +----+                   ...
- *                         |                  |
- *                    partition 1         partition 2
- *                         |                  |
- *                         v                  v
- *
- *                     Decorated           Decorated
- *
- *                   InnerBeamType       InnerBeamType
- * }}}
- */
+  * Beam composed of a stack of smaller beams. The smaller beams are split across two axes: timestamp (time shard
+  * of the data) and partition (shard of the data within one time interval). The stack of beams for a particular
+  * timestamp are created in a coordinated fashion, such that all ClusteredBeams for the same identifier will have
+  * semantically identical stacks. This interaction is mediated through zookeeper. Beam information persists across
+  * ClusteredBeam restarts.
+  *
+  * In the case of Druid, each merged beam corresponds to one segment partition number, and each inner beam corresponds
+  * to either one index task or a set of redundant index tasks.
+  *
+  * {{{
+  *                                            ClusteredBeam
+  *
+  *                                   +-------------+---------------+
+  *               2010-01-02T03:00:00 |                             |   2010-01-02T04:00:00
+  *                                   |                             |
+  *                                   v                             v
+  *
+  *                         +----+ Merged +----+                   ...
+  *                         |                  |
+  *                    partition 1         partition 2
+  *                         |                  |
+  *                         v                  v
+  *
+  *                     Decorated           Decorated
+  *
+  *                   InnerBeamType       InnerBeamType
+  * }}}
+  */
 class ClusteredBeam[EventType: Timestamper, InnerBeamType <: Beam[EventType]](
   zkBasePath: String,
   identifier: String,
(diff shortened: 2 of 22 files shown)


core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -19,22 +19,37 @@
...
+    sendBatch(messages).map(_.size)
+  }
+
+  /**
+    * Request propagation of events. The operation may fail in various ways, which tend to be specific to
+    * the implementing class.
+    * @param messages a batch of messages
+    * @return a bitset containing indexes of messages that sent successfully; other messages were dropped.
+    */
+  def sendBatch(messages: Seq[MessageType]): Future[BitSet]
```
drcrallen: if indicies are returned, parameter should probably be `scala.collection.immutable.Seq`

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -19,22 +19,37 @@
...
 
   /**
-   * Signal that no more events will be sent. Many implementations need this to do proper cleanup.
-   */
+    * Signal that no more events will be sent. Many implementations need this to do proper cleanup.
+    */
   def close(): Future[Unit]
```
drcrallen: Describe the intended use of the future.

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -19,22 +19,37 @@
 package com.metamx.tranquility.beam
 
 import com.twitter.util.Future
+import scala.collection.immutable.BitSet
+import scala.collection.mutable
```
drcrallen: suggest changing to immutable

core/src/main/scala/com/metamx/tranquility/beam/ClusteredBeam.scala
```diff
@@ -347,22 +357,31 @@ class ClusteredBeam[EventType: Timestamper, InnerBeamType <: Beam[EventType]](
...
       }
     }
-    val warmingBeams = Future.collect(for (
-      latestEvent <- grouped.lastOption.map(_._2.maxBy(timestamper(_).millis)).map(timestamper).toList;
-      tbwTimestamp <- toBeWarmed(latestEvent, latestEvent + tuning.warmingPeriod) if tbwTimestamp > latestEvent
-    ) yield {
-      // Create beam asynchronously
-      beam(tbwTimestamp, now)
-    })
+    val latestEventTimestamp: Option[DateTime] = grouped.lastOption map { case (truncatedTimestamp, group) =>
```
drcrallen: This changed from last event per group to last event overall. Can you explain this change a bit?

core/src/main/scala/com/metamx/tranquility/beam/HttpBeam.scala
```diff
@@ -154,13 +157,13 @@ class HttpBeam[A: Timestamper](
                 "eventCount" -> events.size
               )
             )
-            Future.value(0)
+            Future.value(BitSet.empty)
```
drcrallen: (outside scope of this PR) This assumes the the consumer either 

1. Doesn't care about duplicates
2. Has some sort of commit magic that prevents it from committing the batch until the events are reported as reported successfully (say that five times fast!)

Is that assumption documented anywhere?

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -258,14 +254,15 @@ class Tranquilizer[MessageType] private(
...
-          log.debug(s"Sent $n out of ${myBuffer.size} messages. Marking batch dropped.")
-          val e = new MessageDroppedException(s"Batch not complete ($n/${myBuffer.size})")
-          myBuffer.foreach(_.future.setException(e))
+        case Return(bitset) =>
+          log.debug(s"Sent ${bitset.size} out of ${myBuffer.size} messages.")
+          for ((holder, index) <- Beam.index(myBuffer)) {
+            if (bitset.contains(index)) {
+              holder.future.setValue(())
+            } else {
+              holder.future.setException(MessageDroppedException.instance)
```
drcrallen: This no longer contains a message. Is there a reason for that?

core/src/test/scala/com/metamx/tranquility/test/DirectDruidTest.scala
```diff
@@ -104,14 +115,20 @@ class DirectDruidTest
...
+              case Return(()) => Future.value(true)
+              case Throw(e: MessageDroppedException) => Future.value(false)
+              case Throw(e) => Future.exception(e)
+            }
+          })
+          assert(Await.result(eventsSent) === Seq(true, false, true))
           runTestQueriesAndAssertions(broker, timekeeper)
         }
+        catch {
+          case e: Exception =>
```
drcrallen: suggest `NonFatal(e)`

core/src/test/scala/com/metamx/tranquility/test/HashPartitionBeamTest.scala
```diff
@@ -37,11 +38,18 @@ class HashPartitionBeamTest extends FunSuite with Matchers
     override def hashCode(): Int = s(0).toInt
   }
 
-  class TestBeam(callback: TestObject => Unit) extends Beam[TestObject]
+  class DroppingBeam() extends Beam[TestObject]
   {
-    override def propagate(events: Seq[TestObject]): Future[Int] = {
+    override def sendBatch(messages: Seq[TestObject]): Future[BitSet] = Future.value(BitSet.empty)
```
drcrallen: Doesn't a dropped event show up as a 1 in the BitSet?

core/src/test/scala/com/metamx/tranquility/test/TranquilizerTest.scala
```diff
@@ -166,8 +167,8 @@ class TranquilizerTest extends FunSuite with Logging
       newTranquilizer(beam, maxBatchSize, maxPendingBatches, lingerMillis).withFinally(_._1.stop()) {
         case (tranquilizer, desc) =>
           val (acked, dropped, failed) = doSend(tranquilizer, Seq("hey", "__drop__"))
-          assert(acked === 0, "acked (%s)" format desc)
-          assert(dropped === 2, "dropped (%s)" format desc)
+          assert(acked === 1, "acked (%s)" format desc)
+          assert(dropped === 1, "dropped (%s)" format desc)
```
drcrallen: Why did this change?

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: This is only being used to get `mutable.BitSet` so that can be used to merge bitsets

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: I think Seq is ok, there are some times where people want to pass in Buffers they have just created and don't intend to change. If people are going to mutate the Seq after passing it in, they get what's coming to them…

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: sure

core/src/main/scala/com/metamx/tranquility/beam/HttpBeam.scala
gianm: sorta. https://github.com/druid-io/tranquility/blob/master/docs/overview.md#guarantees says,

> If there is an issue that prevents Tranquility from receiving an acknowledgement from the indexing service, it will retry the batch, which can lead to duplicated events.

will address this and one of your other comments by updating the Beam and Tranquilizer interface docs.

gianm: well, just Beam.

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: It's a singleton now, for performance reasons (it's used as sort of a marker for dropped messages, which there can be a lot of if a stream is "catching up" and is behind the windowPeriod)

core/src/test/scala/com/metamx/tranquility/test/DirectDruidTest.scala
gianm: good idea

core/src/test/scala/com/metamx/tranquility/test/HashPartitionBeamTest.scala
gianm: dropped events should not show up in the BitSet, that's how callers know they've been dropped

core/src/test/scala/com/metamx/tranquility/test/TranquilizerTest.scala
gianm: See the removed "This can actually be wrong" comment in Tranquilizer; the return value used to be wrong ("hey" was sent, "__drop__" was dropped, so acked should be 1 and dropped should be 1)

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
gianm: Also the message is less useful now, since the exception to be applied to *all* messages (even the successful ones) and just tell you in the message how many were successful. That was not really that great. Now the exception is only applied to the actually dropped messages.

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -19,22 +19,41 @@
...
+  }
+
+  /**
+    * Request propagation of messages. The returned bitset contains the indexes of messages known to be sent
+    * successfully. Note that for some implementations, it is possible for a message to be sent and for the ack to be
+    * lost (e.g. if the send is occurring over a network).
+    * @param messages a batch of messages
+    * @return a bitset containing indexes of messages that were sent successfully
+    */
+  def sendBatch(messages: Seq[MessageType]): Future[BitSet]
```
drcrallen: This Seq is still mutable, right? meaning there is no hard-guarantee that the BitSet returned by the future can be used on the object that was passed in.

drcrallen: didn't see your comment in the prior comment thread. Since this is not a new interface I don't consider it a blocker, but in general if a sequence is being indexed, it would make sense to me that the passed index must be immutable.

I'm OK as is simply to maintain compat with prior behavior.

gianm: correct, the type system does not guarantee it. But you would have to be crazy to mutate a Seq after passing it to a method like this.

core/src/test/scala/com/metamx/tranquility/test/HashPartitionBeamTest.scala
drcrallen: Yes, I see now. Had something wired backwards in my brain.

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -42,3 +61,27 @@ class DefunctBeamException(s: String, t: Throwable) extends Exception(s, t)
...
 }
+
+object Beam
+{
+  /**
+    * Beam.index(xs) is like xs.zipWithIndex, but always returns an IndexedSeq so lookups by index will be efficient.
+    * @param xs elements
+    * @return indexed seq equivalent to xs.zipWithIndex
+    */
+  def index[A](xs: Seq[A]): IndexedSeq[(A, Int)] = {
```
drcrallen: What happens if `Seq` has more than `Int.MaxValue` items? Both this and the BitSet would break

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
drcrallen: I never did find out the answer to this question yet: Is that style of string formatting lazy? As in, is there a performance difference in wrapping that log.debug statement with a if debug logging enabled, vs just leaving it as is?

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: A lot of things on the Seq trait break down in that case (size, apply, zipWithIndex, and so on) so I think it's fair to assume that Seqs don't have that many items. It would be a really bad idea for someone to create one that did.

drcrallen:
:+1: as is. Minor optional comment about a debug log and performance.

dclim:
Looks good :+1:

PR #79, opened 2016-01-08
Server: Allow configuring DruidBeamConfig.

gianm:


nishantmonu51:
:+1:

dclim:
:+1:

server/src/main/scala/com/metamx/tranquility/server/config.scala
```diff
@@ -69,10 +70,41 @@ object config
...
+
+    @Config(Array("druidBeam.firehoseBufferSize"))
+    def firehoseBufferSize: Int = DruidBeamConfig().firehoseBufferSize
+
     @Config(Array("druid.discovery.curator.path"))
     def discoPath: String = "/druid/discovery"
 
     override def discoAnnounce: Option[DiscoAnnounceConfig] = None
+
+    def druidBeamConfig: DruidBeamConfig = DruidBeamConfig.builder()
```
drcrallen: Do these need validation steps?

gianm: Something crazy will probably happen if you use a negative chunk size or buffer size. If we want to validate that kind of stuff then DruidBeamConfig is probably the right place to do it

PR #80, opened 2016-01-08
Fix doc links, add to "never exiting" section.

gianm:


fjy:
:+1:

rasahner:
+1, but it's always nice if a link to something goes directly to where the information being referenced appears, instead of to the top of a page that you have to read or search through.  This linked page is small enough that it doesn't really matter.

gianm:
@rasahner I can do that, one sec

gianm:
@rasahner updated the first link to include an anchor; I think the second one makes sense linking to the top of the page.

rasahner:
+1

PR #84, opened 2016-01-08
Server: Switch to real arg parsing, and rename Main to ServerMain.

gianm:
Real arg parsing is probably a better idea than reading `args(0)`.

### DIFF
--- a/server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
+++ b/server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
@@ -21,6 +21,7 @@ package com.metamx.tranquility.server.http
 
 import com.metamx.common.lifecycle.Lifecycle
 import com.metamx.common.scala.Abort
+import com.metamx.common.scala.Logging
 import com.metamx.common.scala.Predef._
 import com.metamx.common.scala.Yaml
 import com.metamx.common.scala.collection.implicits._
@@ -39,6 +40,8 @@ import com.metamx.tranquility.server.config.GeneralConfig
 import com.metamx.tranquility.server.config.GlobalConfig
 import com.metamx.tranquility.server.config.GlobalProperties
 import com.metamx.tranquility.tranquilizer.Tranquilizer
+import com.twitter.app.App
+import com.twitter.app.Flag
 import io.druid.segment.realtime.FireDepartment
 import java.io.FileInputStream
 import java.io.InputStream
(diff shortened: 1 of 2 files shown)


fjy:
:+1:

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
```diff
@@ -52,18 +54,25 @@ import org.eclipse.jetty.util.thread.QueuedThreadPool
...
-      new FileInputStream(args(0))
+  private val configFile: Flag[String] = flag(
+    "configFile",
+    s"Path to alternate config file, from working directory (default: $ConfigResource from classpath)"
+  )
+
+  def main(): Unit = {
+    val configInputStream = configFile.get match {
+      case Some(file) =>
+        new FileInputStream(file)
```
drcrallen: Suggest logging information about the fact that you are pulling configs from the file.

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
```diff
@@ -52,18 +54,25 @@ import org.eclipse.jetty.util.thread.QueuedThreadPool
...
+
+      case None =>
+        getClass.getClassLoader.getResourceAsStream(ConfigResource) mapNull {
```
drcrallen: Suggest logging information about the fact that you are pulling configs from the classpath resource.

drcrallen: Also, should you be checking context classloader here first?

gianm: Does that matter for a main?

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
gianm: added logging

server/src/main/scala/com/metamx/tranquility/server/http/ServerMain.scala
drcrallen: Only if you're calling main from a unit test.

gianm: I am not

drcrallen:
:+1:  post Travis

Issue #59, opened 2015-12-04
tranquility-kafka


gianm:
An app that is both a kafka consumer and a tranquility sender. Ideally it should work with only configuration, without users having to write code.

gianm:
fixed by #65

PR #88, opened 2016-01-12
Distribution adjustments.

gianm:
- Add DistributionMain main class
- Move distribution tarball stuff to "distribution" instead of "server"
- Remove dependency on druid-indexing-service (no longer needed)


distribution/src/main/scala/com/metamx/tranquility/distribution/DistributionMain.scala
```diff
@@ -0,0 +1,35 @@
...
+
+object DistributionMain
+{
+  def main(args: Array[String]) {
+    args.headOption match {
+      case Some("server") =>
+        ServerMain.main(args.drop(1))
+
+      case _ =>
+        System.err.println(s"Usage: ${getClass.getCanonicalName} <command> [args]")
```
drcrallen: exit(1)?

gianm: added

drcrallen:
Why is DistributionMain needed separate from ServerMain?

gianm:
@drcrallen #59 will add another main, and sbt's tarball plugin only likes to have one primary main, so this is future proofing.

drcrallen:
:+1:

fjy:
:+1:

PR #90, opened 2016-01-14
Fix dependencies in the sbt build.

gianm:
SBT needs jackson to be forced or else it writes out a bad pom.

See also https://groups.google.com/d/topic/druid-development/CREoK7o1Z-E/discussion

### DIFF
--- a/core/src/test/scala/com/metamx/tranquility/test/DruidBeamTest.scala
+++ b/core/src/test/scala/com/metamx/tranquility/test/DruidBeamTest.scala
@@ -131,7 +131,7 @@ class DruidBeamTest extends FunSuite with Matchers
           }
         }
       }
-    ).forType(classOf[Task])
+    ).withType(classOf[Task])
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/com/metamx/tranquility/test/DruidBeamTest.scala
```diff
@@ -131,7 +131,7 @@ class DruidBeamTest extends FunSuite with Matchers
           }
         }
       }
-    ).forType(classOf[Task])
+    ).withType(classOf[Task])
```
drcrallen: is this an intended change in this PR also?

gianm: yes, `forType` is not in jackson 2.4

xvrl:
:+1:

drcrallen:
:+1: after travis

PR #92, opened 2016-01-15
Apache Flink module

gesundkrank:
This module adds support for Apache Flink. Works and looks a lot like the Spark module. But instead of the `BeamRDD` it contains `BeamSink`.

Had some struggle with the Flink dependency imports since they are not following the Scala naming conventions.

Would be great to get some feedback on this and even greater to get it merge.

### DIFF
--- a/docs/flink.md
+++ b/docs/flink.md
@@ -1,0 +1,53 @@
+# Flink
+
+If you're using Flink to process your event stream, you can use Tranquility's BeamSink to send data to
+Druid. The BeamSink can be added to any DataStream object. The BeamSink requires a BeamFactory to 
+propagate events. You will be overriding the makeBeam() function and within that
+using the DruidBeams builder's "buildBeam()" to build the beam. See the [DruidBeams documentation](druidbeams.md)
+for details about creating beams.
+
+It is recommended that you implement makeBeam as a lazy val so the Beam can be reused.
+
+For example:
+
+```scala
+
+class SimpleEventBeamFactory extends BeamFactory[SimpleEvent]
+{
+
+  lazy val makeBeam: Beam[SimpleEvent] = {
+
+    // Tranquility uses ZooKeeper (through Curator framework) for coordination.
+    val curator = CuratorFrameworkFactory.newClient(
+      "localhost:2181",
+      new BoundedExponentialBackoffRetry(100, 3000, 5)
+    )
+    curator.start()
+
+    val indexService = "druid/overlord" // Your overlord's druid.service, with slashes replaced by colons.
+    val discoveryPath = "/druid/discovery" // Your overlord's druid.discovery.curator.path
+    val dataSource = "foo"
+    val dimensions = IndexedSeq("bar")
+    val aggregators = Seq(new LongSumAggregatorFactory("baz", "baz"))
+
+    DruidBeams
+      .builder((eventMap: SimpleEvent) => new DateTime())
+      .curator(curator)
+      .discoveryPath(discoveryPath)
+      .location(DruidLocation.create(indexService, dataSource))
+      .rollup(DruidRollup(SpecificDruidDimensions(dimensions), aggregators, QueryGranularity.MINUTE))
+      .tuning(
+        ClusteredBeamTuning(
+          segmentGranularity = Granularity.HOUR,
+          windowPeriod = new Period("PT10M"),
+          partitions = 1,
+          replicants = 1
+        )
+      )
+      .buildBeam()
+  }
+}
+
+// now you can create a sink and add it to a given DataStream object.
+dstream.addSink(new BeamSink[SimpleEvent](new SimpleEventBeamFactory(zkConnect)))
+```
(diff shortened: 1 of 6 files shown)


docs/flink.md
```diff
@@ -0,0 +1,52 @@
...
+    val discoveryPath = "/druid/discovery"     // Your overlord's druid.discovery.curator.path
+    val dataSource = "foo"
+    val dimensions = IndexedSeq("bar")
+    val aggregators = Seq(new LongSumAggregatorFactory("baz", "baz"))
+
+    DruidBeams
+      .builder((eventMap: SimpleEvent) => new DateTime())
+      .curator(curator)
+      .discoveryPath(discoveryPath)
+      .location(DruidLocation(indexService, firehosePattern, dataSource))
```
gianm: firehosePattern is no longer needed and most users will find it easier to not worry about it. Could you remove that from the example?

docs/flink.md
```diff
@@ -0,0 +1,52 @@
...
+
+    // Tranquility uses ZooKeeper (through Curator framework) for coordination.
+    val curator = CuratorFrameworkFactory.newClient(
+      "localhost:2181",
+      new BoundedExponentialBackoffRetry(100, 3000, 5)
+    )
+    curator.start()
+
+    val indexService = "druid/overlord" // Your overlord's druid.service, with slashes replaced by colons.
+    val firehosePattern = "druid:firehose:%s" // Make up a service pattern, include %s somewhere in it.
```
gianm: See below on firehosePattern

docs/flink.md
```diff
@@ -0,0 +1,52 @@
...
+    val discoveryPath = "/druid/discovery"     // Your overlord's druid.discovery.curator.path
+    val dataSource = "foo"
+    val dimensions = IndexedSeq("bar")
+    val aggregators = Seq(new LongSumAggregatorFactory("baz", "baz"))
+
+    DruidBeams
+      .builder((eventMap: SimpleEvent) => new DateTime())
```
gianm: Using `new DateTime` for the timestamper is going to lead to mysterious drops (when tranquility thinks an event is ok and reports it sent, but druid silently drops it). I guess you copied this from the spark example, so we should fix both. I don't think that has to happen as part of this PR though.

Filed #93 to remember to do this.

flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.metamx.common.scala.Logging
+import com.metamx.tranquility.tranquilizer.SimpleTranquilizerAdapter
+import org.apache.flink.configuration.Configuration
+import org.apache.flink.streaming.api.functions.sink.RichSinkFunction
+
+/**
+  * This class provides a sink that can propagate any event type to Druid.
+  * @param beamFactory your implementation of [[BeamFactory]].
+  */
+class BeamSink[T](beamFactory: BeamFactory[T])
```
gianm: I don't know enough about flink to know if this implementation makes sense, but as long as flink doesn't need sinks to be thread-safe then it looks good to me.

flink/src/test/scala/com/metamx/tranquility/test/FlinkDruidTest.scala
```diff
@@ -0,0 +1,108 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.metamx.tranquility.test
+
+import java.util.concurrent.TimeUnit
+
+import com.metamx.common.scala.Logging
```
gianm: Could you please import the codestyle jar at https://github.com/druid-io/druid/blob/master/intellij_formatting.jar and optimize the imports? That should split up the grouped `{a, b, c}` imports, which is the style we use for Scala stuff in Druid projects.

gianm:
@gesundkrank thanks for the contribution! Could you please take a look at the comments on the patch, squash your commits (see https://github.com/druid-io/druid/blob/master/CONTRIBUTING.md), and fill out the Druid CLA at http://druid.io/community/cla.html?

docs/flink.md
gesundkrank: Yes. Just copied the Spark example. And I think it makes sense to fix both examples at the same time. Won't fix it in this PR.

flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
gesundkrank: No. Doesn't need to be thread-safe.

gesundkrank:
@gianm Thank you very much for your review! I squashed and commited the fixups. 
My CTO sent you the Druid CLA a few hours ago.

gianm:
:+1: looks good!

@gesundkrank I don't see you in the CLA list- do you fall under a corporate CLA? If so which one?

gesundkrank:
Yes. My company is mbr targeting.

xvrl:
@gesundkrank thank you, received the CLA

PR #96, opened 2016-01-18
Distribution and doc adjustments for releasing tranquility-kafka

gianm:
Also fix flush/counting order in WriterController.


dclim:
:+1:

PR #106, opened 2016-01-27
DruidBeam: Fix exception at TRACE logging level.

gianm:


core/src/main/scala/com/metamx/tranquility/druid/DruidBeam.scala
```diff
@@ -88,7 +88,7 @@ class DruidBeam[A](
       if (log.isTraceEnabled) {
         log.trace(
           "Sending %,d events to task[%s], firehose[%s]: %s",
```
rasahner: %,d really?

gianm: Is that bad?

rasahner: I guess it's a legal format I just wasn't familiar with.

pjain1:
:+1:

rasahner:
+1

PR #109, opened 2016-02-01
tranquility-kafka: switch to kafka 0.8.2.2 for compatibility

### LINKED ISSUE #108: tranquility-kafka does not work with 0.8.x brokers
see https://issues.apache.org/jira/browse/KAFKA-2845

So I think we need to downgrade our kafka lib to work with 0.8.x brokers. It sounds like the older kafka client should work with both 0.8.x and 0.9.x brokers.

dclim:
Tested against Kafka 0.9 brokers and seems to work well. Fixes #108.


gianm:
:+1:

PR #110, opened 2016-02-02
Tranquilizer.flush() to make batch / micro-batch oriented workflows easier

### LINKED ISSUE #85: SimpleTranquilizerAdapter.flush() should force a flush even if lingerMillis has not elapsed
Currently, flush() doesn't actively force Tranquility to send buffered messages to Druid but rather blocks until lingerMillis elapses and the send thread does its thing. This means that the method will block for up to lingerMillis + sendTime when it's called. It would be great if calling flush() would immediately begin flushing the buffer regardless of what lingerMillis is set to.

gianm:
Fixes #85

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
+++ b/core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
@@ -62,25 +62,39 @@ class Tranquilizer[MessageType] private(
 
   case class MessageHolder(message: MessageType, future: Promise[Unit])
 
-  // lock synchronizes access to buffer, bufferStartMillis, pendingBatches.
-  // lock should be notified when waiters might be waiting: buffer becomes non-empty, or pendingBatches decreases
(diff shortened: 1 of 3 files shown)


fjy:
:+1:

core/src/main/scala/com/metamx/tranquility/tranquilizer/Tranquilizer.scala
```diff
@@ -62,13 +62,16 @@ class Tranquilizer[MessageType] private(
 
   case class MessageHolder(message: MessageType, future: Promise[Unit])
 
-  // lock synchronizes access to buffer, bufferStartMillis, pendingBatches.
+  // lock synchronizes access to buffer, bufferStartMillis, pendingBatches, currentBatch*, flushing.
```
xvrl: `*` are cool, but what do they mean

gianm: @xvrl pushed some more comments

dclim:
:+1:

PR #111, opened 2016-02-02
IndexService: close should return a future, and should not attempt to close a client that has not been created yet.

gianm:
I think this should fix the issue reported in https://groups.google.com/d/topic/druid-development/nCMp7JCNhuo/discussion


gianm:
the original poster reports that this patch does indeed fix his issue.

fjy:
:+1:

PR #117, opened 2016-02-05
Console sender.

gianm:
Can be used by piping things to: `bin/tranquility console -configFile conf/core.json.example`


fjy:
:+1:

PR #124, opened 2016-02-18
Add Flink accumulators

gesundkrank:
This PR adds accumulators to Flink BeamSink. They will provide some more insights to the user.

It also updates the readme.

### DIFF
--- a/flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
+++ b/flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
@@ -21,12 +21,16 @@ package com.metamx.tranquility.flink
 import com.metamx.common.scala.Logging
 import com.metamx.tranquility.tranquilizer.MessageDroppedException
 import com.metamx.tranquility.tranquilizer.Tranquilizer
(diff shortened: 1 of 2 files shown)


flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
```diff
@@ -34,16 +38,24 @@ class BeamSink[T](beamFactory: BeamFactory[T])
 {
   var sender: Option[Tranquilizer[T]] = None
 
-  private val exception = new AtomicReference[Throwable]()
+  private val exception       = new AtomicReference[Throwable]()
+  private val sentCounter     = new LongCounter()
```
gianm: In tranquility-kafka and tranquility-server, what you call "sent" and "accepted" are called "received" and "sent". Could you change these to be consistent?

gianm:
looks good after adjusting the naming of the counters.

gesundkrank:
Fixed your comment. Thank you for the review!

gianm:
:+1: Looks good to me!

fjy:
:+1:

PR #125, opened 2016-02-19
More verbose logging in IndexService call failure

drcrallen:


core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
```diff
@@ -119,22 +119,19 @@ class IndexService(
               // Index service can issue redirects temporarily, and HttpClient won't follow them, so treat
               // them as generic errors that can be retried
               throw new IndexServiceTransientException(
-                "Service[%s] temporarily unreachable: %s %s" format
-                  (environment.indexService, code, response.status.reason)
+                s"Service[${environment.indexService}] at ${req.host.getOrElse("unknown")} [${req.path}] temporarily unreachable: $code ${response.status.reason}"
```
gianm: I don't think req.host is going to be set to anything useful (see com.metamx.tranquility.finagle.HttpPost).

drcrallen:
I don't even remember why I had put this in here :-/

PR #126, opened 2016-02-19
Optionally locate Druid tasks through "runningTasks" endpoint on the overlord.

gianm:
Requires druid-io/druid#2419

This is enabled by setting druidBeam.taskLocator=overlord, and works by polling the
overlord in a background thread. Also added a useless setting for
druidBeam.overlordLocator, which might not be useless in the future if we ever have
a way to do that without Curator discovery (perhaps a hard-coded list, or a DNS name).

No integration tests yet, since the Druid code that enables this is not yet in a
Maven release so we cannot pull it in.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/druid/DruidEnvironment.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/DruidEnvironment.scala
@@ -24,7 +24,8 @@ class DruidEnvironment(
 ) extends Equals
 {
   // Replace / with : just like Druid does.
-  val indexService = indexServiceMaybeWithSlashes.replace('/', ':')
+  // See Druid code at https://github.com/druid-io/druid/blob/62ba9ade37f2ab96d3894a0eb622ed5e4f290a03/server/src/main/java/io/druid/curator/discovery/CuratorServiceUtils.java
+  val indexServiceKey: String = indexServiceMaybeWithSlashes.replace('/', ':')
 
   // Sanity check on firehoseServicePattern.
   require(firehoseServicePattern.contains("%s"), "firehoseServicePattern must contain '%s' somewhere")
@@ -34,11 +35,11 @@ class DruidEnvironment(
 
   override def equals(other: Any) = other match {
     case that: DruidEnvironment =>
-      (indexService, firehoseServicePattern) ==(that.indexService, that.firehoseServicePattern)
+      (indexServiceKey, firehoseServicePattern) ==(that.indexServiceKey, that.firehoseServicePattern)
     case _ => false
   }
 
-  override def hashCode = (indexService, firehoseServicePattern).hashCode()
+  override def hashCode = (indexServiceKey, firehoseServicePattern).hashCode()
 }
 
 object DruidEnvironment
--- a/core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
@@ -24,8 +24,10 @@ import com.metamx.common.scala.Predef._
 import com.metamx.common.scala.control._
 import com.metamx.common.scala.exception._
 import com.metamx.common.scala.untyped._
+import com.metamx.tranquility.druid.IndexService.TaskHostPort
 import com.metamx.tranquility.druid.IndexService.TaskId
 import com.metamx.tranquility.finagle._
+import com.twitter.finagle.Addr
 import com.twitter.finagle.Service
 import com.twitter.finagle.http
 import com.twitter.finagle.util.DefaultTimer
@@ -34,12 +36,13 @@ import com.twitter.util.Closable
 import com.twitter.util.Future
 import com.twitter.util.Time
 import com.twitter.util.Timer
+import java.net.InetSocketAddress
 import org.scala_tools.time.Imports._
 
 class IndexService(
   environment: DruidEnvironment,
   config: IndexServiceConfig,
-  finagleRegistry: FinagleRegistry
+  overlordLocator: OverlordLocator
 ) extends Closable
 {
   private implicit val timer: Timer = DefaultTimer.twitter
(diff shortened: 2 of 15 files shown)


gianm:
Note, we have been running this in our demo cluster for a week or so and it has been working for us.

fjy:
:+1:

core/src/main/scala/com/metamx/tranquility/druid/DruidEnvironment.scala
```diff
@@ -24,7 +24,7 @@ class DruidEnvironment(
 ) extends Equals
 {
   // Replace / with : just like Druid does.
-  val indexService = indexServiceMaybeWithSlashes.replace('/', ':')
+  val indexServiceKey: String = indexServiceMaybeWithSlashes.replace('/', ':')
```
drcrallen: Would it be possible to put a comment to where druid does this (with commit hash in url) so that future maintainers can have SOME hope of tracking what to do if this ever changes?

gianm: added a comment.

gianm:
```
[error] (storm/*:update) sbt.ResolveException: download failed: org.apache.storm#storm-core;0.9.3!storm-core.jar
```

seems like some travis goofiness, restarting.

core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
```diff
@@ -87,7 +92,24 @@ class IndexService(
     val statusRequest = HttpGet("/druid/indexer/v1/task/%s/status" format taskId)
     call(statusRequest) map {
       d =>
-        d.get("status") map (sd => IndexStatus.fromString(str(dict(sd)("status")))) getOrElse TaskNotFound
+        dict(d).get("status") map (sd => IndexStatus.fromString(str(dict(sd)("status")))) getOrElse TaskNotFound
+    }
+  }
+
+  def runningTasks(): Future[Map[TaskId, TaskHostPort]] = {
+    val request = HttpGet("/druid/indexer/v1/runningTasks")
```
drcrallen: Is there a way to add a timeout to this?

For example, if an overlord locks up and stops returning results (just holding connection), what is the impact here?

core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
```diff
@@ -166,6 +188,38 @@ object IndexService
...
+        Addr.Neg
+      } else {
+        Addr.Bound(new InetSocketAddress(host, port))
+      }
+    }
+  }
+
+  object TaskHostPort
+  {
+    def unknown = TaskHostPort("", -1)
```
drcrallen: -1 is a valid port for `URL` is this meant to indicate "default" or "invalid"?   If invalid, suggest changing to -2

core/src/main/scala/com/metamx/tranquility/druid/OverlordLocator.scala
```diff
@@ -0,0 +1,68 @@
...
+  val Curator = "curator"
+
+  private[tranquility] def create(
+    kind: String,
+    finagleRegistry: FinagleRegistry,
+    druidEnvironment: DruidEnvironment
+  ): OverlordLocator =
+  {
+    kind match {
+      case `Curator` => new CuratorOverlordLocator(finagleRegistry, druidEnvironment)
```
drcrallen: else throw error?

drcrallen: This part of scala I'm not as familiar with. The resulting return value in the missing case will be an Optional, right?

core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
gianm: There is a timeout (given by finagleHttpTimeout), default is 90 seconds.

gianm: If that happens then the DruidTaskResolver will try again at the next poll period.

core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
gianm: -1 is meant to indicate "unknown" and is also what the overlord would report if the port is unknown.

core/src/main/scala/com/metamx/tranquility/druid/OverlordLocator.scala
gianm: Scala will throw an exception (MatchError- which, despite its name, is actually just a regular Exception) if nothing matches, although a better error is probably a good call just so users see a nice message. I'll change that.

gianm: ok, added a nicer error message.

core/src/main/scala/com/metamx/tranquility/finagle/DruidTaskResolver.scala
```diff
@@ -0,0 +1,144 @@
...
+          nextPoll = 0L
+          tasks = tasks + (taskNumber ->(taskId, new AtomicReference[Addr], updatable))
+          lock.notifyAll()
+        }
+
+        ensureStarted()
+        new Closable
+        {
+          override def close(deadline: Time): Future[Unit] = {
+            lock.synchronized {
```
drcrallen: since deadline is only advisory this can block for as long as it needs to I suppose.

core/src/main/scala/com/metamx/tranquility/finagle/FinagleRegistry.scala
```diff
@@ -31,23 +32,60 @@ import com.twitter.util.Future
...
+      log.info(s"Adding resolver for scheme[${resolver.scheme}].")
+      resolverMap.put(resolver.scheme, resolver)
+    }
+  }
+
+  def schemes: Set[String] = lock.synchronized {
+    resolverMap.keySet
+  }
+
+  private def mkid(scheme: String, name: String): String = s"$scheme!$name"
```
drcrallen: is `!` a forbidden character?

for example, yahoo is commonly represented as "Yahoo!" which would cause problems here.

drcrallen: I see now its a legacy inheritance for naming.

gianm:
@drcrallen any other thoughts?

drcrallen:
:+1:

PR #128, opened 2016-02-22
Cache timestamps when using DruidBeams.fromConfig.

gianm:
~130% speedup in BenchmarkMain (150k/sec -> 350k/sec on my machine.)

Also add a benchmark class (in a separate commit).


fjy:
:+1:

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -39,7 +39,7 @@ import org.joda.time.chrono.ISOChronology
 import org.scala_tools.time.Implicits._
 import scala.util.Random
 
-class DruidBeamMaker[A: Timestamper](
+class DruidBeamMaker[A](
```
dclim: unused import com.metamx.tranquility.typeclass.Timestamper

gianm: removed.

dclim:
:+1:

PR #130, opened 2016-02-23
Tranquilizer blockOnFull option, Server async api

### LINKED ISSUE #87: Server: Allow In Memory Cache of Events
Having an asynchronous endpoint on the server would be great for use cases where you want to post without having to block for task creation, or getting the metrics to a peon.  This would obviously create a few caveats that need to be understood:

1. The client will not receive a positive acknowledgement for how many metrics were delivered to druid.
2. Metrics can be lost if the server jvm dies while metrics are still in the memory cache.

There are cases where we are sending 10K+ metrics every second from a single threaded HTTP post routine, we are bumping into some blocking for these posts on occasion and timing out.  Asynchronous posting would be a nice fix.

gianm:
- Tranquilizer: Add blockOnFull flag to control behavior when outgoing queue is full.
- Server: Add async api.

Fixes #87


fjy:
:+1:

dclim:
:+1:

PR #135, opened 2016-03-08
Update to Flink 1.0.0

gesundkrank:
This PR updates flink-tranquility to use the latest Flink version. 

Fortunately they use proper naming for Scala versions now.


gianm:
:+1:

PR #137, opened 2016-03-09
Fix objectWriter, automatic partitioning for MessageHolder.

### LINKED ISSUE #134: Fix objectWriter, automatic partitioning for MessageHolder
#128 introduced the MessageHolder class, but automatic partitioning and objectWriting don't work for it. This needs to be fixed.

gianm:
Also add a test to DirectDruidTest for MessageHolder-based stacks.

Fixes #134.


dclim:
:+1:

PR #140, opened 2016-03-10
Support all Druid formats by using InputRows internally for default stacks.

### LINKED ISSUE #113: Support for flattenSpec
Tranquility should support passing along the flattenSpecs supported by Druid.

This is particularly useful when you want tranquility-server or tranquility-kafka to handle nested data, but could also be useful for core.

### LINKED ISSUE #119: Support formats other than JSON
Currently we always send data to Druid as JSON. Would be nice to support other formats.

gianm:
It's now possible to make three different kinds of beam stacks with DruidBeams.

1. String/Array[Byte]/ByteBuffer based stacks. This lets you provide raw bytes,
   which will be converted to an InputRow using a Druid StringInputRowParser.
2. Map based stacks. This lets you provide a Map, which will be converted to
   an InputRow using a Druid MapInputRowParser.
3. Classic, custom-type stacks. This gives you full control over everything
   and is the kind of stack that has been with Tranquility since the beginning.

The first kind of stack is new. It's not currently a public API, because I'm
not really sure how best to expose it yet. It's used by Tranquility Kafka and
Tranquility Server.

This requires changing the Beam trait to prefer a new signature,

```
  def sendAll(messages: Seq[MessageType]): Seq[Future[SendResult]]
```

Because the old one (sendBatch) did not allow returning an error for some
messages but not others. This can now happen because the beam stack may
be taking Strings that may turn out to be unparseable. We want to return
errors for those unparseable Strings while still forwarding the rest.

Fixes #113, #119.


xvrl:
@gianm I was under the impression we were trying to reduce the dependance of tranquility on Druid internals, e.g. what was done on the JSON api side. Do you think it would make sense to start moving the pieces that depend on Druid internals into a separate module?

gianm:
@xvrl Most of the new Druid classes being used are from druid-api, so I think that's ok as they are part of Druid's public API rather than Druid internals. I think long run it's probably feasible to limit Tranquility's use of Druid classes to only things in `druid-api` (other than integration tests of course).

The big things *not* in druid-api that we do use are,

- AggregatorFactory (this seems like it would make sense in druid-api, though…)
- FireDepartment (really only using this for deserializing configs, we could get away from it just like we got away from the output classes on the JSON API side)
- QueryGranularity

Although none of those 3 are new in this patch.

gianm:
In practice, AggregatorFactory and QueryGranularity don't change often (& probably would be major releases if they did) so I think if we get rid of FireDepartment, and stick druid-api + AggregatorFactory + QueryGranularity, we'll be in a good place.

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -19,12 +19,13 @@
...
 import com.twitter.util.Future
+import com.twitter.util.Return
+import com.twitter.util.Throw
 import scala.collection.immutable.BitSet
 import scala.collection.mutable
 
 /**
-  * Beams can accept messages and forward them along. The propagate method may throw a DefunctBeamException, which
-  * means the beam should be discarded (after calling close()).
+  * Beams can accept messages and forward them along. The propagate method may throw a
```
dclim: truncated comment

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
```diff
@@ -32,56 +33,101 @@ trait Beam[MessageType]
...
+          case Throw(e) =>
+            throw e
+        }
+      }
+      merged.toImmutable
+    }
+  }
+
+  /**
+    * Request propagation of messages. The returned bitset contains the indexes of messages known to be sent
```
dclim: This one doesn't return a bitset

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: fixed

core/src/main/scala/com/metamx/tranquility/beam/Beam.scala
gianm: fixed

gianm:
@xvrl @dclim do either of you have further comments?

dclim:
If you're okay waiting, I'll take another look at this on Monday - I think I got halfway through and then got distracted by something else.

gianm:
@dclim Monday works for me

dclim:
No more comments from me, :+1:

schmee:
What is the status on this? I'm really looking forward to using FlattenSpec with Tranquility and Kafka 😄

gianm:
@schmee currently waiting on a second reviewer.

@xvrl any further comments?

fjy:
👍

xvrl:
@gianm any performance impact?

gianm:
@xvrl with #167 applied to both master and this branch, this branch is actually a little better.

master + #167:

```
Sent 5000000/5000000 messages in 28511ms (175370 messages/sec) (0 dropped, 0 failed)
Sent 5000000/5000000 messages in 27997ms (178590 messages/sec) (0 dropped, 0 failed)
Sent 5000000/5000000 messages in 28799ms (173617 messages/sec) (0 dropped, 0 failed)
```

this branch + #167:

```
Sent 5000000/5000000 messages in 21733ms (230064 messages/sec) (0 dropped, 0 failed)
Sent 5000000/5000000 messages in 23441ms (213301 messages/sec) (0 dropped, 0 failed)
Sent 5000000/5000000 messages in 21926ms (228039 messages/sec) (0 dropped, 0 failed)
```

gianm:
looking at profiles, the speedup appears to be due to a better default implementation of Partitioner. I don't think that's inherent to the parser changes in this PR – those seem to have little to no performance impact.

PR #141, opened 2016-03-11
Additional logging for non-retryable responses

logarithm:
We faced that sometimes druid returns 4xx error code, but we can't figure out why. It would be nice to have the uri and response body in logs for further investigation of non-retryable error.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
@@ -104,6 +104,7 @@ class IndexService(
 
             case code =>
               // All other responses should not be retried (including non-404 client errors)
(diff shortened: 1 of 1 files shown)


nishantmonu51:
:+1:

gianm:
:+1:

core/src/main/scala/com/metamx/tranquility/druid/IndexService.scala
```diff
@@ -104,6 +104,7 @@ class IndexService(
 
             case code =>
               // All other responses should not be retried (including non-404 client errors)
+              log.error("Non-retryable response for uri[%s] with content: %s" format (req.uri, response.contentString))
```
xvrl: We should make sure the content does not get too long to not pollute the logs

PR #145, opened 2016-03-11
Java 7 test build to avoid issues such as #144

xvrl:


gianm:
:+1:

xvrl:
@gianm looks like jdk7 builds are timing out

gianm:
@xvrl did you see one time out? From what I can see they have not yet started

gianm:
Although that's weird since they've been in the queue for a long time…

gianm:
I'm seeing "Hang tight, the log cannot be shown until the build has started."

given it's been like that for a while, I'm just gonna bounce and see what that does.

xvrl:
@gianm I had rebased on master in case the build got triggered before #144 got merged, so there were multiple builds in the queue. Not sure which one is which though, so we'll just have to wait.

gianm:
@xvrl ah okay, let's just wait then.

gianm:
@xvrl could you try changing this?

```
diff --git a/build.sbt b/build.sbt
index 1d66b95..fae2cdd 100644
--- a/build.sbt
+++ b/build.sbt
@@ -162,7 +162,7 @@ val kafkaTestDependencies = Seq(
 lazy val commonSettings = Seq(
   organization := "io.druid",

-  javaOptions := Seq("-Xms512m", "-Xmx512m"),
+  javaOptions := Seq("-Xms512m", "-Xmx512m", "-XX:MaxPermSize=256M"),

   // Target Java 7
   scalacOptions += "-target:jvm-1.7",
```

xvrl:
@gianm just pushed the change, let's see if it helps

nishantmonu51:
:+1:

PR #146, opened 2016-03-12
Server: Change asynchronous route to an ?async option.

gianm:


docs/server.md
```diff
@@ -35,30 +35,17 @@ This API is synchronous and will block until your messages are either sent to Dr
...
-}
-```
-
-Note that "sent" will always be 0 when using the asynchronous API.
-
-If the server's outgoing message queue fills up, your requests will either block or fail. To control this behavior,
-set the `tranquility.blockOnFull` property to either true or false. If you set this to "false" and your server's
-outgoing queue is full, you will get an HTTP 503.
+If the server's outgoing message queue fills up, your asynchronous requests will either block or fail. To control this
+behavior, set the `tranquility.blockOnFull` property to either true or false. If you set this to "false" and your
```
nishantmonu51: document the default value for tranquility.blockOnFull ?

docs/server.md
```diff
@@ -35,30 +35,17 @@ This API is synchronous and will block until your messages are either sent to Dr
...
+server's outgoing queue is full, you will get an HTTP 503.
 
-With either the synchronous or asynchronous APIs, you can adjust the size of your Tranquility Server's message queues
-through the properties `tranquility.maxBatchSize` and `tranquility.maxPendingBatches`.
+With both the synchronous and asynchronous modes, you can adjust the size of your Server's message queues through
+the properties `tranquility.maxBatchSize` and `tranquility.maxPendingBatches`.
```
nishantmonu51: would be nice to document the defaults if not already done.

gianm: they're documented in configuration.md

gianm: https://github.com/druid-io/tranquility/blob/master/docs/configuration.md

docs/server.md
```diff
@@ -35,30 +35,17 @@ This API is synchronous and will block until your messages are either sent to Dr
 
 If there is an error sending the data, you will get an HTTP error code (4xx or 5xx).
 
-### Asynchronous HTTP API
+### Asynchronous option
 
-Tranquility Server also offers an asynchronous API. The form of the request should be the same as the synchronous API,
-but you should use the URLs `/v1/post-async` or `/v1/post-async/DATASOURCE`.
+Tranquility Server also offers an asynchronous option. Just add "?async=yes" to the URL. Note that "sent" in the
```
nishantmonu51: should we change it to async=true instead of yes ?

nishantmonu51: just to be consistent with other apis having boolean parameters.

gianm: 'true', 'yes' and '1' are all supported, I can change the docs to 'true'

gianm: ok, updated.

dclim:
:+1:

PR #147, opened 2016-03-15
Allow setting lingerMillis to -1 to always wait for complete batches

nishantmonu51:


gianm:
@nishantmonu51 could you add some tests to TranquilizerTest? Perhaps adding `-1` to any of the lingerMillis Seqs that contain `MomentsSoDear`. Those tests should all pass with `-1` too.

gianm:
@nishantmonu51 could you also write in the docs that if you are using linger -1, then you really do need to call `flush` when you're wanting messages to be sent, otherwise they might never get sent.

nishantmonu51:
@gianm done. added test and modified doc.

core/src/test/scala/com/metamx/tranquility/test/TranquilizerTest.scala
```diff
@@ -309,7 +309,7 @@ class TranquilizerTest extends FunSuite with Matchers with Logging
 
 object TranquilizerTest
 {
-  val MomentsSoDear = Seq(525600 * 60000)
+  val MomentsSoDear = Seq(-1)
```
gianm: Could you make this `Seq(525600 * 60000, -1)` (assuming this does not cause problems with the tests)?

Reason being, the tests for "long lingering" should work with both a big value and with -1.

nishantmonu51: makes sense, done.

gianm:
:+1:

xvrl:
:+1: after travis

PR #152, opened 2016-03-18
DruidBeamMaker: Use availabilityGroup no longer tied to baseFirehoseId.

gianm:
This, combined with the runningTasks taskLocator, means that the cycleBucket
hack will eventually not be used anymore.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
@@ -151,14 +151,15 @@ class DruidBeamMaker[A](
       beamTuning.segmentGranularity.widen(interval) == interval,
       "Interval does not match segmentGranularity[%s]: %s" format(beamTuning.segmentGranularity, interval)
     )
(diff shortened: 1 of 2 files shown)


nishantmonu51:
can you explain more on why was the cycleBucket hack initially introduced ? was it creating many buckets in Zookeeper ?

gianm:
it was to avoid creating too many znodes in service discovery. Curator does not remove old znodes, and if you have a service discovery key unique to each task, eventually you have thousands of stale znodes.

nishantmonu51:
after this change even with runningTasks tasklocator, druid will still announce the firehose in service discovery. This will cause issues for the existing setups.

gianm:
@nishantmonu51 which issues are you thinking of? I didn't change the firehose announcement path in this patch, to try to avoid any potential issues like that. the idea is that after this code is rolled out, the firehose will still be announcing in service discovery at the cycleBucket path, but nothing will be using it. eventually we can disable that if we want.

nishantmonu51:
nm, i thought you removed the cycleBucket hack based on the PR description. 
looked at the code more closely now, seems good to me. 
:+1:

dclim:
:+1:

Issue #158, opened 2016-04-23
[Spark Streaming] Background Send Thread Not Exit


du00cs:
I use tranquility with spark streaming for some while and I find a big problem that will make it impossible to run long enough. A [Google Group thread](https://groups.google.com/forum/#!topic/druid-user/zJXLPpDLV70) has mentioned it in Feb. 

With the streaming job running, 
```bash
jstack 69069 | grep Tranquilizer-BackgroundSend- | wc -l
```
shows that growing number of threads in waiting.
```
// com.metamx.tranquility.tranquilizer.Tranquilizer: L101

val sendThread = loggingThread {
    try {
      while (!Thread.currentThread().isInterrupted) {
        val (exIndex, exBuffer) = lock.synchronized {
          while (
            buffer.isEmpty ||
              pendingBatches.size >= maxPendingBatches ||
              (!flushing && lingerMillis > 0 && System.currentTimeMillis() < bufferStartMillis + lingerMillis)
          ) {
            if (lingerMillis == 0 || bufferStartMillis == 0) {
              lock.wait() // here
```
More and more threads in memory will at last make the streaming job exceed the memory limit and killed by yarn.

The `stop` method of `Tranquilizer` can not be called at all, for it will make the sender in next stages sent nothing! 

My solution is rather simple, manually stop the sender thread. And after doing this, the daemon threads will not grow any more.

If you need some reproducible cases, I can attach them here. I will create a PR soon later.

du00cs:
@fjy Debug for some while and I locate an issue with it. This will improve the stability and I will test the throughput later. Hope it works too.

plelevier:
Need a fix for this too

du00cs:
@plelevier Resolved in master. You can try it now

PR #159, opened 2016-04-23
[Spark Streaming] Fix memory overhead of growing number of background send threads

du00cs:
See issue #158 

Simply interrupt the thread after sending will fix the issue. I think this problem make spark streaming with tranquility nearly unusable. I suffer from this for weeks...


gianm:
Hmm, that this fix works that indicates we are creating WAY too many Tranquilizers. There should only be one… ideally we could fix this by sharing Tranquilizers rather than by interrupting the ones that are getting leaked.

du00cs:
I traced for some while and find that tranquilizer is the same instance of tranquilizer, but each streaming batch (stage) will call the `start()` method to create new thread. The problem is here. I have been using the ugly interrupted one for some while, seems no problem.

gianm:
hey @du00cs, thanks for looking into this and tracking down the problem. Could you please take a look at #168 and see if that works for you? I think that should fix this too but hopefully more nicely, without creating the extraneous threads in the first place.

du00cs:
@gianm Thanks for fixing. I will test it.

du00cs:
@gianm It works.

gianm:
@du00cs thanks for testing! If you don't mind then I will close this PR and get in that other one.

du00cs:
@gianm 👍

PR #169, opened 2016-05-05
Allow for overlapping tasks with tweak to cycleBucket

boneill42:
We have a window period that causes tasks to overlap.  This is a small change to the algorithm which calculates the bucket to allow for such cases.  

Feel free to close this PR if you don't want to allow for this, or have already solved it with other enhancements.


gianm:
Thanks for the patch @boneill42. I tweaked it a bit (added tests, changed the modulus to 3, removed an extraneous shell script) and submitted it as #171. You should still get author credit though.

FWIW the real fix for this issue is setting `druidBeam.taskLocator = overlord`, which will work starting in Druid 0.9.1. At that point this cycleBucket stuff won't be needed.

gianm:
Do you mind trying that out and seeing if it works for you?

fjy:
@boneill42 This fix should be a part of Tranquility 0.8.0. Feel free to close this issue if you verify that it works correctly.

boneill42:
Verified.

gianm:
thanks!

Issue #53, opened 2015-11-23
Tranquility causes data loss due to re-use of finangle service name


rohitkochar:
i have hit a bug where tranquility will cause data loss due to re-use of the service name of the finangle service ,for certain combination of segment granularity and window period.

Here are the details of my setup:
 Segment granularity:Minute
 Window period = 120 minutes
 datatsource name = ds1

In case of segment granularity is minute,tranquility picks only the minute part of the task interval along with the replicant number and partition number to form the name of the finangle service.

For an interval of [2015-11-19T07:00 - 2015-11-19T07:01] and for replicant number 0 and partition number 0 the service name would be "<firehose pattern>ds1-00-0000-0000" where firehose pattern is a static value.

Tranquility uses this service name to query the zoo keeper to  identify the <peon host> and <peon port> details where it can send data and in scenarios where same service name is used to connect to more than 1 firehose than it will get multiple destinations to for a service.

In the scenario given above same service name would be used to connect to task running at 7:00 and task running at 6:00 and since window period is 2 hours when a DruidBeam is constructed to post events to task of interval 7:00 it will get <host:port> details of two different peon since 6:00 task would still be alive.
If a finagle client is created with more than 1 destination then it starts sending data to them in round robin way in order to do load balancing hence causing half the data of 7:00 to go to task 6:00 which eventually drops it and hence data loss happens.

I have verified that in my setup i see no data loss in scenarios where window period is less than 60 minutes and segment granularity is minute whereas i see almost half data getting lost in the scenario explained above.

gianm:
@rohitkochar, yes, this is a limitation of the way the service names get generated, when your windowPeriod is > your segmentGranularity. This is one of the reasons we recommend having windowPeriod < segmentGranularity (see the bottom of: http://druid.io/docs/latest/ingestion/realtime-ingestion.html).

I think you could work around this by modifying how the service names get generated, to have it cycle through a few more ids (maybe include hour mod 6?) before repeating the cycle. Do you think that sort of change would be helpful to you?

Long term the best solution is for the overlord to tell tranquility where it can find a task, so we don't need to use service discovery like this.

rohitkochar:
What would be the side effects of using the entire interval range as part of the service name??
Is it that we will create too many Z nodes in zookeeper(one for each service) which won't get deleted?

Lets say i want the window period of 2 hours,than my segment granularity has to be DAY.
In that case how many realtime task would be launched per day?
[...]

PR #171, opened 2016-05-06
Add some additional hourly cycling to minutely cycleBuckets

gianm:
Partially addresses #53 using a slightly modified version of the patch in #169 (% 3 instead of % 2, and updated tests). Should also cover #72.

The real fix for this issue is setting `druidBeam.taskLocator = overlord`, which will work starting in Druid 0.9.1. Until then, this helps work around issues with longer windowPeriods on sub-hour granularity tasks.

This should be backwards compatible, because existing firehoseIds are retrieved from the serialized dicts. `generateFirehoseId` is only called for new tasks.


fjy:
👍

dclim:
👍

PR #173, opened 2016-05-23
Add config for sending data to Druid using Smile.

gianm:
Enabled by setting `serialization.format: smile`. This was already possible with custom objectWriters, but the config makes it easier to use.


fjy:
👍

PR #176, opened 2016-05-31
Lengthen DirectDruidTest timeouts.

gianm:
Hopefully help out slower CI boxes.


gianm:
hmm, things still failed, bummer. I wonder if we need an EVEN LONGER TIMEOUT?

Or if there's some other resource issue we're hitting in travis… the direct druid tests are pretty resource hungry.

PR #178, opened 2016-06-16
Add Second Granularity to DruidBeamMaker

andremleblanc:
Allows segmentGranularity to be set to SECOND. Previously, if SECOND was specified, a failed to create merged beam error would be generated caused by java.lang.IllegalArgumentException: No gross firehose id hack for granularity[SECOND].


core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -243,6 +243,7 @@ object DruidBeamMaker
     val tsUtc = new DateTime(ts.millis, ISOChronology.getInstanceUTC)
 
     val cycleBucket = segmentGranularity match {
+      case Granularity.SECOND => tsUtc.hourOfDay().get % 3 * 60 + tsUtc.minuteOfHour().get // 10800 buckets
```
gianm: This should involve "secondOfMinute" somewhere in it, otherwise two seconds for the same minute would get the same bucket (which we want to avoid- tasks for nearby but distinct time ranges should have different buckets)

andremleblanc: Ahh, I was being a dummy...I'll update it after I put the kids to bed 😄

core/src/main/scala/com/metamx/tranquility/druid/DruidBeamMaker.scala
```diff
@@ -243,6 +243,7 @@ object DruidBeamMaker
     val tsUtc = new DateTime(ts.millis, ISOChronology.getInstanceUTC)
 
     val cycleBucket = segmentGranularity match {
+      case Granularity.SECOND => (tsUtc.minuteOfHour().get * 60 + tsUtc.secondOfMinute().get) % 900 // 900 buckets
```
andremleblanc: Without increasing the number of characters allowed for this portion of the base firehouse ID, 999 buckets is the most we can create. I chose 900 buckets because that is an even distribution of seconds over an hour (3600 / 4) while remaining under the 3 digit cap. What I wrote will sequentially fill 900 buckets and then reset after the 900th bucket every 15 mins.

gianm: Cool, looks good!

gianm:
👍

andremleblanc:
Travis failed, but it looks like the test it failed on (DirectDruidTest) is intermittently failing, so I don't think any action is required from me. If my assumption is incorrect, I'll be happy to dig in.

nishantmonu51:
:+1:

nishantmonu51:
reopening to trigger another travis run.

nishantmonu51:
@andremleblanc could you please sign the CLA here - http://druid.io/community/cla.html

andremleblanc:
@nishantmonu51 absolutely, just signed it.

gianm:
thanks for the contribution @andremleblanc!

Issue #161, opened 2016-04-26
tranquility-flink supports only one sink operation?


donovankim:
if two BeamSink exists in one flink program, 
` getRuntimeContext.addAccumulator("Druid: Messages received", receivedCounter)`
occured error. because accumulator aleady exists.

gesundkrank:
@donovankim Thanks for reporting the error and sorry for the late reply.

I opened #181 that will fix it.

PR #180, opened 2016-06-21
Added option to Flink sink to throw an exception on dropped messages.

gesundkrank:
Also updated Flink dep to latest version.


flink/src/main/scala/com/metamx/tranquility/flink/BeamSink.scala
```diff
@@ -32,8 +32,9 @@ import org.apache.flink.streaming.api.functions.sink.RichSinkFunction
   * This class provides a sink that can propagate any event type to Druid.
   *
   * @param beamFactory your implementation of [[BeamFactory]].
+  * @param throwOnDrop throws an exception if set to true and a message gets dropped.
   */
-class BeamSink[T](beamFactory: BeamFactory[T])
+class BeamSink[T](beamFactory: BeamFactory[T], throwOnDrop: Boolean = false)
```
gianm: This is called `reportDropsAsExceptions` in other modules, could you use that name here?

gesundkrank: Sure. Committed a fixup.

gianm:
👍  thanks!

btw, do you have any thoughts on #161?

gesundkrank:
Thank you! 

Haven't seen the issue. Thanks for the hint. Will open a PR that fixes it, soon.

gianm:
@gesundkrank sweet, thanks!

Issue #191, opened 2016-08-31
How to build a tranquility distribution


iAlaska:
Hi, I am new to SBT. 
And now ,i want to build tranquility from source to generate a distribution Tarball.

I've execute `sbt +package` but i could not find the tar.gz file in the target directory.

thanks a lot!

du00cs:
I just solved the problem today. `sbt 'project distribution' bundle:dist` , if it does not work, 

```bash
sbt
> compile
> project distribution
> bundle:dist
```

this should work.

iAlaska:
Great thanks.  I found the solution in druid-usergoup yesterday.
```
sbt 'project distribution' universal:packageZipTarball
```

PR #192, opened 2016-09-01
[DOC] Add build distribution doc

du00cs:
See #191 , I think someone else confused too.


gianm:
Thanks @du00cs for the contribution. Sorry it took so long to get to it. I usually use a slightly different command to build a distribution (like the one on http://druid.io/downloads.html), so I raised #214 including that one.

PR #206, opened 2017-01-11
Upgrading finagle version to 6.35.0

revathy144:
Upgrading finagle version to 6.35.0. We use tranquility for druid in our organization Tapad. Tranquility uses 6.31.0 version of finagle which is not compatible to some of our services. So we request to upgrade finagle to 6.35.0. 

This would also require scala-utils, which is dependency for tranquility-core to upgrade its finagle version to 6.35.0. I have created a pull-request requesting them to upgrade their project's finagle version to 6.35.0


logarithm:
@revathy144 could you please fix travis-ci tests? Also you can try excluding finagle version for scala-utils and tranquility in your project.

drcrallen:
Related comment https://github.com/druid-io/tranquility/pull/220#issuecomment-294508028

logarithm:
@revathy144 Could you please close this PR since requested changes were merged in separate PR. Thanks!

PR #220, opened 2017-04-14
Update MMX libraries and replace scala_tools.time

logarithm:
* Update MMX libraries versions
* Replaced scala_tools.time by nscala_time since scala_tools.time doesn't support Scala 2.12 and nscala_time is being actively maintained


logarithm:
Seems like new finagle 6.35.0 is failing for scala 2.10 version: https://groups.google.com/forum/#!msg/finaglers/Q86RGwoHHmQ/QKNNGQ--CwAJ

We can't drop scala 2.10 support, at least because we still on old samza version.

logarithm:
I found what causes OOM exception in tranquility tests. It's because of this line: https://github.com/druid-io/tranquility/blob/1905a5027896889ddd0af750f916079dde3bdf13/core/src/main/scala/com/metamx/tranquility/finagle/FinagleRegistry.scala#L89-L89. I think we have next options:
1. Make failFast `true` by default in `FinagleRegistryConfig`. It affects clients who rely on default behaviour.
2. Make failFast `true` in tests. This helps to make tests pass, but clients, who use scala 2.10 and `finagleEnableFailFast` `false` will fail with OOM.
3. Drop Scala 2.10 support. This leads to update samza and kafka versions.
4. Build our own finagle version with fix.

I don't think first option might be applied here, but I've put it in the list just to show that we have it. For me personally third option looks as the best choise. @gianm what do you think about it?

logarithm:
Created PR https://github.com/druid-io/tranquility/pull/221 to remove scala 2.10 first

drcrallen:
Can we depend on old finagle for scala 2.10?

logarithm:
If we want to update to the latest `scala-util` we can't use old finagle because it will be incompatible with `scala-util` code.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: scala
 scala:
-   - 2.10.5
-   - 2.11.7
+   - 2.10.6
+   - 2.11.8
```
gianm: Seems like this conflicts with #221, maybe merge the PRs into a single one? Or did you want one of them reviewed first?

build.sbt
```diff
@@ -184,6 +184,8 @@ lazy val commonSettings = Seq(
 
   homepage := Some(url("https://github.com/druid-io/tranquility")),
 
+  resolvers += "sigar" at "https://repository.jboss.org/nexus/content/repositories/thirdparty-uploads/",
```
gianm: Why is this necessary? I thought sigar was marked as optional.

gianm:
@logarithm hmm, I think dropping 2.10 support is fine. Are you going to have to upgrade Samza first? If so could you confirm that a new setup (new Scala + new Tranquility + new Samza) is working at your site before we do an OSS tranquility release?

build.sbt
logarithm: I'll remove it with new `scala-util` version. In one of previous versions of `server-metrics` library we removed `optional` for sigar. Now we made sigar optional again.

.travis.yml
logarithm: As I described in my previous comments scala 2.10 has weird bug with Streams and tranquility fails with OOM. I think it's better to merge #221 first and then merge this PR.

gianm: Sounds good, let's do #221 first

build.sbt
gianm: Ah, okay. Are you planning to publish that before this PR is merged?

logarithm:
@gianm thanks for the review! I'll run tests on real data in the next couple of days and let you know about results.

gianm:
thanks @logarithm !

build.sbt
logarithm: Yes, I've just published it and going to update this PR soon.

logarithm:
@gianm I've tested tranquility with scala 2.11, samza 0.12, kafka 0.10.1.1 and other updated libraries. It works in our test cluster. So, I think we ready to merge #221 first and then this PR.

PR #221, opened 2017-04-17
Remove scala 2.10 support

logarithm:
* Removed scala 2.10 support
* Updated samza and kafka versions


samza/src/test/scala/com/metamx/tranquility/test/SamzaDruidTest.scala
```diff
@@ -114,6 +118,70 @@ class SamzaDruidTestInputSystemFactory extends SystemFactory
   }
 }
 
+class SamzaDruidCoordinatorSystemFactory extends SystemFactory
```
gianm: What's this used for?

build.sbt
```diff
@@ -1,6 +1,4 @@
-scalaVersion := "2.10.6"
-
-crossScalaVersions := Seq("2.10.6", "2.11.8")
+scalaVersion in ThisBuild := "2.11.8"
```
gianm: Want to add a 2.12.x build too? Or is that a bad idea for some reason?

samza/src/test/scala/com/metamx/tranquility/test/SamzaDruidTest.scala
logarithm: New samza uses `System` for storing configs. This is test `SystemFactory` which allow us test new samza without starting test kafka broker.

build.sbt
logarithm: Samza doesn't have build for 2.12 yet, they merged PR with samza support but haven't released it yet (https://github.com/apache/samza/pull/82). We can add 2.12 support but build samza for 2.11 only (as we do now for 2.10).

gianm: Adding 2.12 support but only building samza for 2.11 sounds good to me.

logarithm: Cool! I'll add it then.

logarithm: @gianm Since build for scala 2.12 will lead to update of finagle/twitter version (builds for scala 2.12 starts from `6.41.0` version), hence we have to update mmx scala-util library. Do you think in this case it's better to close this PR and remove/add scala versions in #220 ?

gianm: Hmm, I would also be fine with not supporting Scala 2.12 just yet, if it's a bunch of extra work. We should do it eventually but I guess it doesn't have to be now. Nobody has asked for it yet that I am aware of.

I'll leave it up to you. If you _do_ want to add 2.12 then yeah, I think it makes sense to do that all as part of #220.

gianm: Hmm, I would also be fine with not supporting Scala 2.12 just yet, if it's a bunch of extra work. We should do it eventually but I guess it doesn't have to be now. Nobody has asked for it yet that I am aware of.

I'll leave it up to you. If you _do_ want to add 2.12 then yeah, I think it makes sense to do that all as part of #220.

logarithm: I've checked libraries which tranquility uses and it's not trivial to add scala 2.12 support right now (for example scala 2.12 support in flink is coming soon, spark version should be updated also and so on) . I would suggest dropping scala 2.10 support for now and then add 2.12 support later by another PR.

logarithm:
Removed oraclejdk7 from travis build. Tranquility can still be compiled with target=1.7, but to run tests it requires JDK8 because 0.12 samza version uses Java 8 default methods.

gianm:
@logarithm is this still needed since #220 is merged?

logarithm:
@gianm no, I think this PR might be closed.

PR #223, opened 2017-05-09
Exclude deps with *GPL licenses.

gianm:
Swap findbugs (similar to https://github.com/druid-io/druid/issues/3866) and exclude the mysql connector from scala-util (which we aren't using).


fjy:
👍

PR #228, opened 2017-05-30
Thread-safe samza BeamProducer

logarithm:
Since we use samza 0.12 which might work in multi-threading environment druid producer should be thread safe. `Tranquilizer`'s methods `send` and `flush` already thread safe, but `BeamProducer` was not.

This patch addresses it and makes `BeamProducer` thread safe.


samza/src/main/scala/com/metamx/tranquility/samza/BeamProducer.scala
```diff
@@ -55,16 +56,18 @@ class BeamProducer(
     throwOnError
   )
 
+  // PermanentMap supports only get by key, but we need to be able to iterate through all streams
+  private val streams = mutable.Set[String]()
```
gianm: This isn't thread safe, does it need to be?

samza/src/main/scala/com/metamx/tranquility/samza/BeamProducer.scala
```diff
@@ -76,6 +79,7 @@ class BeamProducer(
     val sender = senders.getOrElseUpdate(
       streamName, {
         log.info("Creating beam for stream[%s.%s].", systemName, streamName)
+        streams += streamName
```
gianm: This won't be atomic with the update to `senders`, does it need to be?

logarithm:
@gianm Updated with double-checked locking and ConcurrentHashMap

gianm:
LGTM, thanks @logarithm

gianm:
LGTM, thanks @logarithm

PR #248, opened 2017-08-03
extra checks to prevent Tranquilizer of being stuck

isimonenko:
Tranquilizer relies on Beam chain to respect contract and return future for each message sent with `sendAll`. However it have no control on what Beam implementation is used and will be stuck if `futureResults` is empty. 
We can't check that futures in response are in order but we can at least check the size match and the collection is not empty and fail fast.


isimonenko:
@gianm Could you please take a look?

PR #257, opened 2017-11-15
expose sslContext and prefer tlsPort if present

pjain1:
Fixes https://github.com/druid-io/tranquility/issues/255
Depends on https://github.com/metamx/scala-util/pull/37

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/finagle/FinagleRegistry.scala
+++ b/core/src/main/scala/com/metamx/tranquility/finagle/FinagleRegistry.scala
@@ -78,7 +78,7 @@ class FinagleRegistry(
         }
       )
     }
-    val client = ClientBuilder()
+    val clientBuilder = ClientBuilder()
       .name(id)
       .codec(http.Http())
       .dest(Name.Bound(resolver.bind(name), id))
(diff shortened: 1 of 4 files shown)


himanshug:
:+1: let us update scala-util dependency version too as soon as https://github.com/metamx/scala-util/pull/37 is done and released.

pjain1:
Travis will fail till scala-util dependency is updated after metamx/scala-util#37 is released

pjain1:
@gianm changed to use `Option`, please verify if it is indeed what you expected. I have tested it locally with https enabled on both Overlord and MiddleManager.

core/src/main/scala/com/metamx/tranquility/finagle/FinagleRegistry.scala
```diff
@@ -87,7 +87,8 @@ class FinagleRegistry(
       .logger(FinagleLogger)
       .daemon(true)
       .failFast(config.finagleEnableFailFast)
-      .build()
+    val client = if(config.sslContextOption.isDefined) clientBuilder.tls(config.sslContextOption.get).build() else clientBuilder.build()
```
gianm: More scala-y:

```scala
val client = config.sslContextOption match {
  case Some(sslContext) => clientBuilder.tls(sslContext).build()
  case None => clientBuilder.build()
}
```

Or even:

```scala
config.sslContextOption.foreach(sslContext => clientBuilder.tls(sslContext))
val client = clientBuilder.build()
```

pjain1:
Trying to fix issues with with overlord based task discovery

pjain1:
@gianm can you have a look again

himanshug:
:+1:

PR #266, opened 2018-01-24
Change default host/port for DruidNode and FlinkBeam

amalakar:
This PR contains few fixes I had to add to make it work with flink (1.4) & protobuf 3.5.1 :
- Druid comes with an old version of guava (16.x) and newer version of guava has `HostAndPort.getHostText`  method removed as per [api docs](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/net/HostAndPort.html#getHostText--
)
```
Deprecated. Use getHost() instead. This method is scheduled for removal in Guava 22.0.
```
 
I had a 22.0 of guava in my classpath so it was throwing 
```
java.lang.NoSuchMethodError: com.google.common.net.HostAndPort.getHostText()Ljava/lang/String;
```
Passing null for both host and port doesn't exercise the above code path. Thus avoids calling the deprecated method.

- While extending the flink `BeamFactory` in java I had trouble as it was complaining about abstract method `tranquilty()` not being defined. I am not very familiar with making scala & Java interoperate so not sure if that is the best way to resolve that, ended up moving that to the BeamSink instead.


mistercrunch:
@gianm can you take a look? thx!

gianm:
LGTM - thanks for the contrib @amalakar!

amalakar:
Thank you @gianm !

PR #277, opened 2018-06-25
Support Basic HTTP auth with druid, TLS support for server

jon-wei:
This PR adds support for Basic HTTP authentication when Tranquility communicates with Druid.

The PR also adds configuration options that allow Tranquility Server to accept requests with TLS and communicate with a TLS-enabled Druid cluster.

### DIFF
--- a/core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
+++ b/core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
@@ -359,7 +359,13 @@ object DruidBeams extends Logging
       .partitions(config.propertiesBasedConfig.taskPartitions)
       .replicants(config.propertiesBasedConfig.taskReplicants)
       .druidBeamConfig(config.propertiesBasedConfig.druidBeamConfig)
-      .generalConfig(config.propertiesBasedConfig)
+      .basicAuthUser(config.propertiesBasedConfig.basicAuthUser)
+      .basicAuthPass(config.propertiesBasedConfig.basicAuthPass)
+      .tlsEnable(config.propertiesBasedConfig.tlsEnable)
+      .tlsProtocol(config.propertiesBasedConfig.tlsProtocol)
+      .tlsTrustStoreAlgorithm(config.propertiesBasedConfig.tlsTrustStoreAlgorithm)
+      .tlsTrustStoreType(config.propertiesBasedConfig.tlsTrustStoreType)
+      .tlsTrustStorePassword(config.propertiesBasedConfig.tlsTrustStorePassword)
   }
 
   /**
@@ -435,15 +441,6 @@ object DruidBeams extends Logging
     require(fireDepartment.getIOConfig.getFirehoseFactoryV2 == null, "Expected null 'firehoseV2'")
     require(fireDepartment.getIOConfig.getPlumberSchool == null, "Expected null 'plumber'")
     fireDepartment
-  }
-
-  /**
-    * Return a new injected SSLContext
-    */
-  private[tranquility] def makeSSLContext(): Option[SSLContext] =
-  {
-    val injectedSSLContext = DruidGuicer.Default.get[SSLContext];
-    Option.apply(injectedSSLContext)
   }
 
   class Builder[InputType, EventType] private[tranquility](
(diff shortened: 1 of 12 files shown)


jon-wei:
marking as WIP, there are some unit test failures related to passing configuration, but don't have time this week to look further

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
```diff
@@ -432,6 +437,15 @@ object DruidBeams
     fireDepartment
   }
 
+  /**
+    * Return a new injected SSLContext
+    */
+  private[tranquility] def makeSSLContext(): Option[SSLContext] =
+  {
+    val injectedSSLContext = DruidGuicer.Default.get[SSLContext];
+    Option.apply(injectedSSLContext)
```
gianm: `Some(injectedSSLContext)` is more idiomatic Scala, if you know that this will be non-null, or `Option(injectedSSLContext)` if it might be null.

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
```diff
@@ -725,6 +739,17 @@ object DruidBeams
       )
     }
 
+    /**
+      * Provide tunings for communication with Druid tasks. Optional, see [[DruidBeamConfig]] for defaults.
+      *
+      * @param beamConfig beam config tunings
+      * @return new builder
+      */
+    def generalConfig(generalConfig: PropertiesBasedConfig) = {
```
gianm: This is a bit weird. Generally the Builder is set up to take each config option separately. The idea is that if you're using Tranquility programmatically (i.e. using Core or Storm or Samza), each Builder method call is equivalent to specifying a particular property with a property-based config (i.e. using Server or Kafka).

Any reason it has to be done this way, or would it be possible to split this up into a Builder method for each option?

core/src/main/scala/com/metamx/tranquility/security/BasicAuthClientMaker.scala
```diff
@@ -0,0 +1,41 @@
...
+
+object BasicAuthClientMaker extends Logging
+{
+  def wrapBaseClient(
+    baseClient: Service[http.Request, http.Response],
+    basicAuthUser: String,
+    basicAuthPass: String
+  ): Service[http.Request, http.Response] = {
+    if (basicAuthUser.length > 0 && basicAuthPass.length > 0) {
+      log.info("Using BasicAuth.Client...")
```
gianm: Minor, but, IMO log messages look better without the dots.

core/src/main/scala/com/metamx/tranquility/security/BasicAuthClientMaker.scala
```diff
@@ -0,0 +1,41 @@
...
+      BasicAuth.client(basicAuthUser, basicAuthPass).andThen(baseClient)
+    } else {
+      log.info("Using client without authentication..")
```
gianm: Same here.

core/src/main/scala/com/metamx/tranquility/security/SSLContextMaker.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import com.metamx.common.scala.Logging
+import com.metamx.tranquility.config.PropertiesBasedConfig
+
+object SSLContextMaker extends Logging
+{
+  def createSSLContextOption(generalConfig: PropertiesBasedConfig): Option[SSLContext] = {
+    if (!generalConfig.tlsEnable) {
+      log.info("TLS is not enabled, skipping SSLContext creation.")
+      return Option.empty[SSLContext]
```
gianm: It's more idiomatic to return `None` here. And it's even more idiomatic to avoid `return` in favor of more functional-ish looking code (where the entire function is effectively a single expression).

core/src/main/scala/com/metamx/tranquility/security/SSLContextMaker.scala
```diff
@@ -0,0 +1,45 @@
...
+    catch {
+      case ex@(_: CertificateException |
+               _: KeyManagementException |
+               _: IOException |
+               _: KeyStoreException |
+               _: NoSuchAlgorithmException) =>
+        throw new RuntimeException(ex)
+    }
+
+    Option.apply(sslContext)
```
gianm: `Option(sslContext)` is more idiomatic, if sslContext might be nullable, or `Some(sslContext)` if you know it's not nullable.

server/src/main/scala/com/metamx/tranquility/server/PropertiesBasedServerConfig.scala
```diff
@@ -19,6 +19,10 @@
 
 package com.metamx.tranquility.server
 
+import java.security.KeyStore
+import java.util
```
gianm: This makes the code look a bit weird, imo. I usually prefer `import java.{util => ju}` and then using `ju.List` etc.

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
jon-wei: This function wasn't actually used (I originally planned on injecting the SSLContext using Druid's ssl client extension but the current tranquility druid version and latest druid version are rather far apart so I abandoned that)

core/src/main/scala/com/metamx/tranquility/druid/DruidBeams.scala
jon-wei: Split this into a builder method for each option

core/src/main/scala/com/metamx/tranquility/security/BasicAuthClientMaker.scala
jon-wei: Removed dots

core/src/main/scala/com/metamx/tranquility/security/BasicAuthClientMaker.scala
jon-wei: Removed dots

core/src/main/scala/com/metamx/tranquility/security/SSLContextMaker.scala
jon-wei: Changed this to avoid `return` and replaced this with `None`

core/src/main/scala/com/metamx/tranquility/security/SSLContextMaker.scala
jon-wei: Changed this to `Some(sslContext)`

server/src/main/scala/com/metamx/tranquility/server/PropertiesBasedServerConfig.scala
jon-wei: Changed this to `ju.List`

PR #294, opened 2019-04-10
Handling exceptions while closing BeamSink

KKcorps:
Currently Flink is not able to close the BeamSink properly when Tranquility isn't initialised properly in open method. The initialisation can be prohibited due to NoClassDefFoundError in usual scenarios due to dependency conflicts. As a result Flink throws 

> ERROR org.apache.flink.streaming.runtime.tasks.StreamTask           - Error during disposal of stream operator.
> java.util.NoSuchElementException: None.get

This pull request catches this error and simply logs it.


KKcorps:
Hi
In absence of this exception handling the tranquilizer is not able to close properly and leads to unexpected failures when run inside docker or yarn

